PR #4, opened 2023-10-15
Added new regression and normalization options

ooples:
Added new regression and normalization options

Moved all parameter options to a new regression options class

Updated examples to reflect new changes

Added new examples


LucasDLopes (approved):
Test approval, Didn't see anything wrong with code.

PR #27, opened 2023-10-30
Parallel Processing Logic

jdehorty:
Main point is the example of parallel processing logic, which I mentioned on our call. Feel free to scavenge whatever your need from this PR and modify the formatting to whatever your preference is.

### DIFF
--- a/src/Regression/MultipleRegression.cs
+++ b/src/Regression/MultipleRegression.cs
@@ -1,113 +1,106 @@
-﻿namespace AiDotNet.Regression;
+﻿using System;
+using System.Threading.Tasks;
+using MathNet.Numerics.LinearAlgebra;
 
-public sealed class MultipleRegression : IRegression<double[], double>
+namespace AiDotNet.Regression
 {
-    private double YIntercept { get; set; }
-    private double[] Coefficients { get; set; } = Array.Empty<double>();
-    private MultipleRegressionOptions RegressionOptions { get; }
-
-    public double[] Predictions { get; private set; }
-    public IMetrics Metrics { get; private set; }
-
-    /// <summary>
-    /// Performs multiple regression on the provided inputs and outputs.
-    /// This handles all of the steps needed to create a trained ai model including training, normalizing, splitting, and transforming the data.
-    /// </summary>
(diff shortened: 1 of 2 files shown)


src/Regression/MultipleRegression.cs
```diff
@@ -1,113 +1,137 @@
...
+        // Transform method
+        internal override double[] Transform(double[][] inputs)
+        {
+            // Initialize predictions array and store lengths for optimization
+            var predictions = new double[inputs[0].Length];
+            var inputsLength = inputs.Length;
+            var inputs0Length = inputs[0].Length;
+
+            // Use parallel processing for large data sets
+            Parallel.For(0, inputsLength, i =>
```
ooples, line 126: I understand the need for Parallel processing for large datasets but this will actually create a performance bottleneck for small to average sized datasets so we should dynamically use Parallel.For when the inputs Length is a certain minimum size

src/Regression/MultipleRegression.cs
```diff
@@ -1,113 +1,137 @@
...
         }
 
-        return predictions;
+        // Transform method
+        internal override double[] Transform(double[][] inputs)
+        {
+            // Initialize predictions array and store lengths for optimization
+            var predictions = new double[inputs[0].Length];
+            var inputsLength = inputs.Length;
+            var inputs0Length = inputs[0].Length;
```
ooples, line 123: It is a minor issue but you are referencing the same inputs[0].Length on line 121 and 123 so you should move up your variable and use that variable for line 121

src/Regression/MultipleRegression.cs
```diff
@@ -1,113 +1,137 @@
-﻿namespace AiDotNet.Regression;
+﻿using System;
+using System.Threading.Tasks;
+using MathNet.Numerics.LinearAlgebra;
```
ooples, line 3: It is another minor issue but we moved all using statements to Helpers/UsingsHelper.cs so if you can please move the necessary using statements there.

ooples (changes requested):
A general comment but I wasn't able to find any changes for what you talked about with adding customization options for stuff like Matrix Decomposition, etc so I feel like I'm just being blind

PR #695, opened 2026-01-04
perf: use cached GPU buffers and lower slow test thresholds

ooples:
## Summary
- Fix critical GPU buffer caching bug where `TryGetCachedBuffer()` existed but was never called
- Lower slow test reporting thresholds to catch 20-30 second tests

### GPU Buffer Caching (Critical Performance Fix)

**Root Cause**: `TryGetCachedBuffer()` method existed but was never called in GPU operations, causing weights/biases to be re-uploaded from CPU to GPU on every forward pass.

**Fixed Methods**:
- `FusedLinear`: Now checks cache for weights and bias tensors
- `FusedConv2D`: Now checks cache for kernel tensors  
- `FusedBatchNorm`: Now checks cache for gamma, beta, and running statistics

**Implementation**:
- Check `TryGetCachedBuffer()` before allocating new GPU buffers
- Skip float conversion when using cached buffers (saves CPU work)
- Added proper `finally` blocks to dispose only owned buffers (not cached ones)

**Impact**: For MoE (Mixture of Experts) inference with 8 experts, this eliminates ~10.3GB of wasted memory bandwidth per expert generation by reusing persistent tensors registered via `RegisterPersistentTensor()`.

### Slow Test Thresholds

Updated CI reporting thresholds to catch optimization candidates earlier:
- **CRITICAL**: Changed from >60s to >30s (15x over 2s target)
- **WARNING**: Changed from 30-60s to 10-30s (5-15x over target)  
- **NEW**: Added OPTIMIZE category for 2-10s tests (over 2s target)

Target per-test runtime: 2 seconds

## Test plan
- [x] Verify builds pass for net8.0
- [x] Verify builds pass for net471
- [ ] CI tests should pass
- [ ] GPU operations should use cached buffers when available

### DIFF
--- a/src/AiDotNet.Tensors/AiDotNet.Tensors.csproj
+++ b/src/AiDotNet.Tensors/AiDotNet.Tensors.csproj
@@ -7,7 +7,7 @@
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <Version>1.0.0-preview</Version>
     <Title>AiDotNet.Tensors</Title>
-    <Description>High-performance tensor operations with SIMD acceleration for .NET. Provides Vector, Matrix, and Tensor types with hardware-accelerated operations (Sin, Cos, Exp, Log, etc.) that work across .NET Framework 4.7.1 and .NET 8.0. Supports any numeric type via generic INumericOperations interface. GPU acceleration available.</Description>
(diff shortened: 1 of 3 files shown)


ooples:
@coderabbitai full review

Issue #1015, opened 2026-03-20
perf: JIT compiler integration for zero-allocation diffusion model forward passes


ooples:
## Goal

Complete the AiDotNet JIT compiler integration with TensorWorkspace to enable zero-allocation, fused-kernel execution for diffusion models and all neural networks. The JIT compiler already has 8 optimization passes, 170+ IR operations, and 20+ fused operations but is not wired into the forward pass of any model.

### Graph Capture for Neural Networks
- [ ] **LayerBase graph capture mode**: Add a mode where Forward() records operations to a computation graph instead of executing them. First call captures, subsequent calls execute compiled graph.
- [ ] **NeuralNetworkBase.CompileForward()**: Method that exports the full forward pass as a computation graph, runs JIT compilation, and stores the compiled function.
- [ ] **UNetNoisePredictor compiled forward**: Wire ForwardUNet to use compiled graph when available.
- [ ] **DiffusionResBlock graph export**: ExportComputationGraph currently delegates to conv1 only. Needs to export the full GroupNorm->SiLU->Conv->TimeEmbed->GroupNorm->SiLU->Conv->Residual chain.

### Missing IR Operations for Diffusion
- [ ] **GroupNormOp**: GroupNorm is in every DiffusionResBlock but has no IR operation. Cannot be optimized or fused by the JIT compiler.
- [ ] **FusedGroupNormActivationOp**: Fuse GroupNorm + SiLU (the most repeated pattern in diffusion UNets). Eliminates one 41MB intermediate tensor per ResBlock.
- [ ] **FusedConv2DBiasActivationOp**: Maps to IEngine.FusedConv2D. The diffusion UNet uses Conv+Bias+Identity (not Conv+BatchNorm), so we need this distinct from FusedConvBatchNormActivationOp.

### Missing Fusion Patterns
- [ ] **GroupNorm + SiLU**: Pattern 11 in OperationFusionPass
- [ ] **Conv2D + Bias + SiLU**: Pattern 12 (uses FusedConv2D from IEngine)
- [ ] **GroupNorm + SiLU + Conv2D**: Pattern 13 (3-op fusion for the full DiffusionResBlock half-block)
- [ ] **Residual Add + GroupNorm**: Pattern 14 (fuses the skip connection add with the next norm)

### Missing Optimization Passes
[...]

PR #1142, opened 2026-04-14
feat: wire JIT compilation through AiModelBuilder

ooples:
## Summary

Plugs the `AiDotNet.Tensors` compiled-plan infrastructure into `AiModelBuilder` so a single `.ConfigureJitCompilation()` call on the builder delivers auto-compiled forward-pass replay to every neural network. The Tensors package has shipped `CompiledModelCache`, `CompiledInferencePlan`, `CompiledTapeTrainingStep`, and the TensorCodec optimization pipeline (dataflow fusion / attention fusion / algebraic backward / CSE / BLAS batching) — but AiDotNet had no builder-level surface to configure it. The builder's YAML schema already documented a `jitCompilation` section but nothing populated the knob.

### 1. `NeuralNetworkBase.Predict` virtual default routes through `PredictCompiled`

`Predict` changes from `abstract` to `virtual` with a default that wraps `PredictCompiled(input)` in `NoGradScope<T>`. All 246 existing subclass overrides continue unchanged (their `override` still binds), so this is a non-breaking API evolution. Subclasses with trivial `Predict(x) => Forward(x)` bodies can drop their override to pick up compiled inference.

Cache invalidation plumbed into `InvalidateParameterCountCache` and `RestoreOriginalParameters` so a lazy-init resize or structure change drops any plan captured against the old layer graph.

### 2. `AiModelBuilder.ConfigureJitCompilation` + `JitCompilationConfig` + result persistence

- New `src/Configuration/JitCompilationConfig.cs` — POCO mirroring the `TensorCodecOptions` phase flags (EnableCompilation, EnableDataflowFusion, EnableAttentionFusion, EnableConstantFolding, EnableForwardCSE, EnableBlasBatch, etc.) plus a `ThrowOnFailure` strict-mode switch. Presets: `Default`, `Aggressive`, `Disabled`. `ApplyToTensorCodec()` projects onto and installs the Tensors-package options on the current thread.

- `AiModelBuilder.ConfigureJitCompilation(JitCompilationConfig?)` — follows the `ConfigureMixedPrecision` / `ConfigureReasoning` pattern. Config applied at the top of `BuildAsync` so every subsequent step (training, quantization calibration, cross-validation, RL rollouts) runs under the configured codec flags.

- Config persisted onto `AiModelResultOptions` and `AiModelResult`, and re-applied on every `Predict` call. `TensorCodecOptions.Current` is `[ThreadStatic]` — without this bridge, cross-thread inference (request pools, ASP.NET handlers) would see library defaults instead of the builder's flags.

- YAML wiring: added `JitCompilation` property on `YamlModelConfig` and `YamlConfigApplier` binding. The source generator's `existingSections` set already listed `"JitCompilation"` so the hand-written property takes precedence cleanly.

- `IAiModelBuilder` interface updated so every implementation gets the new method signature.

### 3. Populate `JitCompiledFunction` so JIT reaches `Predict`-override models

The previous two commits only set `TensorCodecOptions` flags. That's enough when a model uses the base `NeuralNetworkBase.Predict` default (from commit #1). It's NOT enough when a concrete model overrides `Predict` — and most do: VGG, CNN, FeedForwardNeuralNetwork, etc. return `Forward(input)` directly, bypassing `PredictCompiled`. Only AutoTracer (Layer 2, capped at 128 recorded ops) benefits them — too short for deep transformers.

This commit populates the `JitCompiledFunction` slot on `AiModelResult` at build time via a new `BuildCompiledPredictFunction` helper. The helper wraps the model's `Predict` in a `CompiledModelCache` regardless of how the concrete override is written. `AiModelResult.Predict` already checks `JitCompiledFunction` at line 1877 and routes to the compiled path when set.

Key implementation details:
- Gated on `_jitCompilationConfig.Enabled` and `model is NeuralNetworkBase<T>`. Diffusion models (DiffusionModelBase does not extend NeuralNetworkBase) and non-neural models (regression, trees) return `null` — JIT doesn't engage and the result's Predict stays on its existing path. _Diffusion coverage is tracked as a structural refactor in a follow-up PR._
- One `CompiledModelCache` per wrapper. `GetOrCompileInference(Tensor, Action)` traces on miss, rebinds input data on hit — no per-call allocation.
- **Nested-GraphMode defense**: during tracing, temporarily set `TensorCodecOptions.EnableCompilation = false` so any call into `NeuralNetworkBase.PredictCompiled` from within the model's own `Predict` falls through to `PredictEager` instead of opening a second, conflicting GraphMode scope. Without this guard the inner compile drops the outer trace's ops and replay produces wrong results.
- Applies JIT config to the current thread on every call.
- Try/catch with `ThrowOnFailure` behavior: by default, compilation failures fall back silently to eager. Setting `JitCompilationConfig.ThrowOnFailure = true` propagates them — use in tests.

## Known limitations (documented inline)

- **Diffusion models bypass**: `DiffusionModelBase` doesn't extend `NeuralNetworkBase`, so the wrapper returns `null` for diffusion. Closing this requires a structural change (`CompiledModelHost<T>` composition component) that's scoped as its own follow-up PR.
- **Silent wrong-results risk**: If a covered model has non-Engine tensor access (direct `AsSpan` writes, scalar control flow) in its forward, trace succeeds but replay reads stale data. Try/catch catches hard failures; silent divergence won't trigger. Users opting into JIT should flip `ThrowOnFailure = true` in tests and diff compiled vs eager outputs.

## Build verification

## Usage

```csharp
// Enable with defaults (silent fallback to eager on failure)
var result = await new AiModelBuilder<float, Tensor<float>, Tensor<float>>()
    .ConfigureModel(myModel)
    .ConfigureJitCompilation()
    .BuildAsync();

// Aggressive (all fusion/CSE passes on)
await builder.ConfigureJitCompilation(JitCompilationConfig.Aggressive).BuildAsync();

// Strict mode for tests
var cfg = JitCompilationConfig.Default;
cfg.ThrowOnFailure = true;
await builder.ConfigureJitCompilation(cfg).BuildAsync();
```

## Summary by CodeRabbit

* **New Features**
  * Configurable JIT compilation with a fluent builder option and three presets (Default, Aggressive, Disabled). Settings persist with built models, survive cloning/serialization, and can be loaded from YAML.

* **Bug Fixes / Improvements**
  * Safer JIT: per-shape failure short-circuiting, automatic fallbacks to eager execution, configurable failure behavior, and invalidation of compiled caches when model structure changes. JIT settings are reapplied per prediction for consistent behavior across threads.

### DIFF
--- a/src/AiModelBuilder.cs
+++ b/src/AiModelBuilder.cs
@@ -1156,6 +1156,125 @@ public partial class AiModelBuilder<T, TInput, TOutput> : IAiModelBuilder<T, TIn
     /// </remarks>
     public Task<AiModelResult<T, TInput, TOutput>> BuildAsync() => BuildAsync(CancellationToken.None);
 
+    /// <summary>
+    /// Wraps a trained model's <c>Predict</c> in a <c>CompiledModelCache</c>-backed
+    /// function matching the <c>JitCompiledFunction</c> shape expected by
+    /// <see cref="AiModelResult{T, TInput, TOutput}"/>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This is how JIT compilation reaches every model whose <c>Predict</c> override
+    /// bypasses the base class. The wrapper traces the model's forward pass under
+    /// GraphMode on the first call at each input shape, compiles it into a flat
+    /// replay plan, and re-executes the plan for subsequent calls.
+    /// </para>
+    /// <para>
+    /// Nested-GraphMode defense: during tracing the wrapper temporarily sets
+    /// <c>TensorCodecOptions.EnableCompilation = false</c> so any call into
+    /// <c>NeuralNetworkBase{T}.PredictCompiled</c> from within the model's own
+    /// <c>Predict</c> falls through to <c>PredictEager</c> instead of opening a
+    /// second, conflicting GraphMode scope.
+    /// </para>
+    /// <para>
+    /// Scope: only wraps <see cref="NeuralNetworks.NeuralNetworkBase{T}"/>
+    /// descendants. Diffusion models (extend <c>DiffusionModelBase</c>, not
+    /// <c>NeuralNetworkBase</c>) and non-neural models (regression, trees) return
+    /// <c>null</c> so the result's Predict path stays on its existing code path.
+    /// Multi-step inference (autoregressive text, RNN unroll) that relies on
+    /// per-step scalar control flow will either trace correctly if the loop is
+    /// unrolled at compile time, or fall through to eager via the try/catch —
+    /// if it traces correctly but produces wrong results on replay (a real risk
+    /// for models with non-Engine tensor access), set
+    /// <see cref="AiDotNet.Configuration.JitCompilationConfig.ThrowOnFailure"/>
+    /// in tests to catch silent divergence.
+    /// </para>
+    /// </remarks>
+    private Func<Tensor<T>[], Tensor<T>[]>? BuildCompiledPredictFunction(
+        IFullModel<T, TInput, TOutput>? model)
+    {
+        if (model is null) return null;
+        if (_jitCompilationConfig is null || !_jitCompilationConfig.Enabled) return null;
+
+        // Only NeuralNetworkBase<T> models go through the Engine-op forward pass
+        // that GraphMode can record. Regression/tree models use scalar math that
+        // wouldn't benefit from graph compilation even if it were traceable.
+        if (model is not NeuralNetworks.NeuralNetworkBase<T> nnModel) return null;
+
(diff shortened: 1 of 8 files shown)


ooples:
Refs #1015 — this PR adds the `ConfigureJitCompilation` builder surface for the JIT compiler integration tracked in issue #1015. Contributes to the "NeuralNetworkBase.CompileForward()" checklist item by exposing the existing Tensors-side `CompiledModelCache`/`CompiledInferencePlan` infrastructure through a fluent builder entry point. Does NOT close #1015 — that issue also covers IR operations (GroupNormOp, FusedGroupNormActivationOp), fusion patterns (Patterns 11-14), optimization passes (memory planning, tile scheduling), and benchmark infrastructure — all tracked separately.

PR #1143, opened 2026-04-15
refactor: CompiledModelHost foundation + Dispose cascade across NN/Diffusion

ooples:
## Summary

Extracts the compiled-inference cache that lived as an ad-hoc field on `NeuralNetworkBase` into a sealed composable component (`CompiledModelHost<T>`), then wires the same component into `NoisePredictorBase` and adds a Dispose cascade hook on `DiffusionModelBase`. This is the foundation for closing the structural gaps that have caused several rounds of perf-feature whack-a-mole — one attachment point for compilation, Dispose, and every future feature (AOT plan serialization, CUDA Graph capture, persistent autotune, plan stitching).

Plan reference: `~/.claude/plans/lucky-waddling-knuth.md` Part B.

### 1. `refactor: extract compile cache into composable CompiledModelHost<T>`

- New `src/NeuralNetworks/CompiledModelHost.cs` — sealed component, ~140 LOC. Predict(input, structureVersion, eagerForward) traces on miss, replays on hit, falls back to eager on disable/failure. Auto-invalidates on `structureVersion` mismatch — callers bump the version when their layer graph mutates and stale plans get dropped before the next compile.
- `NeuralNetworkBase` replaces ad-hoc `_compiledInferenceCache` field with `_compileHost`. `PredictCompiled` becomes a one-line delegation. `Dispose(bool)` now cascades to every layer implementing `IDisposable` so pool-rented weight tensors return to the allocator before GC. `try/catch(ObjectDisposedException)` for shared-layer graphs.

### 2. `refactor: wire NoisePredictorBase through CompiledModelHost + Dispose`

- `NoisePredictorBase` implements `IDisposable`. No existing predictor declared its own Dispose — pure addition.
- Owns its own `CompiledModelHost<T>` for per-step Forward compilation. The diffusion 50-step denoising loop becomes one trace + 49 replays instead of 50 eager full-graph executions, when concrete predictors opt in via the new protected `PredictCompiled(input, eagerFallback)` helper.
- `protected virtual EnumerateLayers()` — concrete predictors override to expose private layer fields (`_blocks`, `_timeEmbed1`, `_patchEmbed`, etc.) for Dispose cascade. Default returns empty so this is non-breaking.
- `InvalidateCompiledPlans()` for concrete predictors to call after lazy-init expansion or weight swap.

### 3. `refactor: DiffusionModelBase opt-in Dispose cascade via EnumerateDisposableComponents`

[...]

PR #1144, opened 2026-04-15
perf: fused fwd+bwd+optimizer as one compiled kernel in TrainWithTape

ooples:
## Summary

Wires `ICompiledTrainingPlan.ConfigureOptimizer` through `NeuralNetworkBase.TrainWithTape` so forward + backward + parameter update run as ONE compiled replay kernel instead of three separate passes with materialized gradients between them. Beats PyTorch's separate `optimizer.step()` for plain Adam/AdamW/SGD training loops — the Adam moment update happens inside the plan's flat delegate array, SIMD-accelerated, zero allocation.

Opt-in via the same `TensorCodecOptions.EnableCompilation` gate the other compile features use, so a single `ConfigureJitCompilation()` on the builder (landed in PR #1142) unlocks both inference AND training compilation.

Plan reference: `~/.claude/plans/lucky-waddling-knuth.md` Part C PR 2.

## Tensors-side primitives (already shipped)

- `ICompiledTrainingPlan<T>.ConfigureOptimizer(OptimizerType, lr, betas, eps, wd)` — installs a closure over the per-parameter m/v buffers and delegates to `FusedOptimizer.AdamUpdateSimd` / `AdamWUpdateSimd` / `SgdUpdateSimd`.
- `Plan.Step()` now runs fwd → bwd → optimizer update as one kernel when `ConfigureOptimizer` has been called.

### `src/Training/CompiledTapeTrainingStep.cs` — new `TryStepWithFusedOptimizer`

Mirrors the existing `Step()` but replaces the hardcoded `UpdateParametersSGD` with `plan.ConfigureOptimizer + plan.Step`. Tracks plan identity via `[ThreadStatic] _lastConfiguredPlan` so `ConfigureOptimizer` is called EXACTLY ONCE per compiled plan — re-calling would reset Adam's m/v buffers and destroy training. Returns `bool` + `out T lossValue` so the caller can distinguish success from fallback-needed.

### `src/NeuralNetworks/NeuralNetworkBase.cs` — `TrainWithTape` fast-path preamble

`TryTrainWithFusedOptimizer()` is called first. If it returns `true` the method returns without running the eager tape path. Otherwise eager tape runs as before.

`TryMapToFusedOptimizerConfig` gates on:
- `TensorCodecOptions.Current.EnableCompilation == true`
- `typeof(T) == typeof(float)` (Tensors-side limitation — fused kernels operate on `float*` directly)
- Optimizer is `AdamOptimizer` / `AdamWOptimizer` / `StochasticGradientDescentOptimizer` with `UseAdaptiveLearningRate == false` (adaptive rates mutate hyperparameters between steps, which would reset the compiled plan's closure-captured m/v buffers)
- At least one `ITrainableLayer` participates
- `LossFunction` derives from `LossFunctionBase<T>`

Any failure short-circuits to `false`, eager tape runs instead — no correctness risk for users outside the supported configuration.

## Behavior unchanged for

- double-precision models (fall through — Tensors doesn't fuse fp64 yet)
- users with custom optimizers (Nadam, RMSprop, Lion, etc.)
- users with learning-rate schedulers or adaptive rates
- users who haven't enabled compilation

## Expected impact on supported configurations

~10-20% training throughput improvement on Adam-trained float models, primarily from eliminating the materialized gradient tensor and the separate `optimizer.Step` call's dispatch overhead. Most notable on small-batch regimes where dispatch dominates.

## Build verification

## Stacking

Independent of #1142 (JIT wiring) and #1143 (CompiledModelHost foundation) — all branch off master. They'll rebase cleanly regardless of merge order. This PR is most valuable after #1142 since `ConfigureJitCompilation()` is the one-line enable.

## Summary by CodeRabbit

* **New Features**
  * Added an accelerated fused/compiled training path for float models with supported optimizers (SGD, Adam, AdamW). Training now tries the fused fast-path first and falls back to the standard tape-based flow if unavailable.
  * Early optimizer resolution before training and deduplicated parameter handling to avoid double-updating shared weights.

* **Bug Fixes**
  * Fused-path state is fully cleared on reset and when invalidated so safe recompilation can occur.
  * Fused execution now disables itself after unsupported or failed attempts and falls back reliably, with safer detection of optimizer hyperparameter drift.

### DIFF
--- a/src/NeuralNetworks/NeuralNetworkBase.cs
+++ b/src/NeuralNetworks/NeuralNetworkBase.cs
@@ -2456,6 +2456,19 @@ public abstract class NeuralNetworkBase<T> : INeuralNetworkModel<T>, IInterpreta
     protected void TrainWithTape(Tensor<T> input, Tensor<T> expected,
         IGradientBasedOptimizer<T, Tensor<T>, Tensor<T>>? optimizer = null)
     {
+        var resolvedOptimizer = optimizer ?? GetOrCreateBaseOptimizer();
+
+        // Fused-compiled fast path: forward + backward + parameter update all
+        // in one compiled kernel, SIMD-accelerated, zero materialized gradient
+        // tensors between backward and the optimizer step. Engages when the
+        // optimizer is a plain Adam/AdamW/SGD with constant hyperparameters
+        // AND the numeric type is float (Tensors-side fused optimizer kernels
+        // operate on float* directly). Returns false if any condition isn't
+        // met or tracing fails; we then fall through to the eager tape path
+        // below with behavior unchanged.
+        if (TryTrainWithFusedOptimizer(input, expected, resolvedOptimizer))
+            return;
+
         // Initialize parameter buffer — replaces layer tensors with buffer-backed views.
         // try/finally MUST cover this so views are restored even if setup throws.
         var initialParams = Training.TapeTrainingStep<T>.CollectParameters(Layers, _parameterBuffer is null ? -1 : _layerStructureVersion);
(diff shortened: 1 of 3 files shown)


ooples:
Refs #1015 — this PR wires fused forward+backward+optimizer into `TrainWithTape` using the `ICompiledTrainingPlan.ConfigureOptimizer` API from the Tensors package. Contributes to the JIT integration tracked in #1015 specifically for the training path (complementing #1142's inference-time JIT wiring). Does NOT close #1015 — IR operations, fusion patterns, and benchmark infrastructure are tracked separately.

Issue #1136, opened 2026-04-14
perf: fix 5 cancelled CI jobs — Diffusion models OOM/timeout from eager weight allocation


ooples:
## Summary

5 GitHub Actions jobs hit the 45-minute workflow timeout and get cancelled on every PR run (baseline: PR #1135 run [24398739627](https://github.com/ooples/AiDotNet/actions/runs/24398739627)):

- `Tests (net10.0) - ModelFamily - Diffusion A-I`
- `Tests (net10.0) - ModelFamily - Diffusion J-R`
- `Tests (net10.0) - ModelFamily - Diffusion S-Z`
- `Tests (net10.0) - ModelFamily - Generated Layers`
- `Tests (net10.0) - ModelFamily - NeuralNetworks`
- `Tests (net10.0) - Unit - 03 Diffusion/Encoding`
- `Tests (net10.0) - Unit - 08e NN-Remaining (catch-all)`

Across the five cancelled jobs: **~370 per-test timeouts** (xunit 60s/120s `[Fact(Timeout=...)]`) and **~950 `OutOfMemoryException`**. Only 18-638 tests finish before the wall-clock kill.

Tests purposely use production defaults to catch real performance bugs, so the fix must be in the model code — not in the tests.

This issue covers **PR 1 of a series**: the Diffusion jobs. Follow-up issues/PRs will clean up NeuralNetworks, Generated Layers, and the two Unit shards using the same infrastructure.

## Root causes (verified against code)

Every Diffusion OOM traces through the same path:

```
System.OutOfMemoryException
  at TensorAllocator.Rent[T](Int32[] shape)
  at DenseLayer`1..ctor (DenseLayer.cs:367)
  at DiTNoisePredictor`1.InitializeLayers (DiTNoisePredictor.cs:334 or 338)
  at DiTNoisePredictor`1.EnsureLayersInitialized (DiTNoisePredictor.cs:301)
  at DiTNoisePredictor`1.PredictNoise (442)  OR  .GetParameters (966)
  at <Model>.GetParameters
  at DiffusionModelTestBase.Parameters_ShouldBeNonEmpty (332)
```

DiT-XL defaults: `hiddenSize = 1152`, `numLayers = 28`, `mlpRatio = 4.0` → ~4 GB of eagerly-allocated weight tensors **per model instance**. Rented tensors are never returned to the pool, so sequential tests across 255 diffusion models stack up on 16 GB Windows runners.

**Five contributing factors:**

[...]

PR #1146, opened 2026-04-15
perf: wire gradient checkpointing into ForwardForTraining via existing TrainingMemoryConfig

ooples:
## Summary

`TrainingMemoryConfig.UseGradientCheckpointing` and `CheckpointEveryNLayers` have lived on the builder via `ConfigureMemoryManagement` for some time — documented but never wired to actual forward-pass logic. Setting the flags had no effect. This PR closes that gap.

Plan reference: `~/.claude/plans/lucky-waddling-knuth.md` Part C PR 5.

## Design note

No new `Configure*Checkpointing` method. The library already has:
- `ConfigureCheckpointManager(ICheckpointManager)` — model-state persistence between runs
- `ConfigurePipelineParallelism(..., ActivationCheckpointConfig, ...)` — distributed pipeline-parallel training

Adding a third would be conceptual clutter. Instead this uses the EXISTING `ConfigureMemoryManagement(TrainingMemoryConfig)` method whose POCO already has the right fields — just wire them through.

### `src/NeuralNetworks/NeuralNetworkBase.cs`

- `ForwardForTraining` checks a per-instance segment size. When > 0 and layer count exceeds it, routes through `AiDotNet.Tensors.Engines.Autodiff.GradientCheckpointing<T>.Checkpoint` — runs segments under `NoGradScope` (skips activation storage), registers one checkpoint op per segment whose backward recomputes forward-with-tape and threads gradients through. Memory: O(layers) → O(sqrt(layers)) at ~33% backward-compute overhead.
- New internal `SetGradientCheckpointingSegmentSize(int)` lets the builder push configured values.

### `src/AiModelBuilder.cs`

- `BuildAsync` reads `_memoryConfig.UseGradientCheckpointing` and when true, computes the effective segment size (`CheckpointEveryNLayers`, or auto-sqrt(layers) when ≤ 0) and pushes it onto the model **before** any training inside the Build runs (quantization calibration, cross-validation, fine-tuning).

## User-facing API

Exactly the existing `ConfigureMemoryManagement` with no new surface:

```csharp
// Manual tuning
builder.ConfigureMemoryManagement(new TrainingMemoryConfig {
    UseGradientCheckpointing = true,
    CheckpointEveryNLayers = 4   // or 0 for auto sqrt(N)
});

// Or via preset
builder.ConfigureMemoryManagement(TrainingMemoryConfig.ForTransformers());
```

## Primitive

Uses the **public** `AiDotNet.Tensors.Engines.Autodiff.GradientCheckpointing<T>.Checkpoint(functions, input, segmentSize)` that's already shipped in the Tensors package. No Tensors-side changes needed. (The separate `CompiledTrainingPlan.EnableCheckpointing` on the internal sealed class — filed as ooples/AiDotNet.Tensors#165 — would add compiled-plan checkpointing as a follow-up; this PR handles the eager tape path.)

## Build verification

Behavior unchanged for users who don't set `UseGradientCheckpointing = true`.

## Reference

Chen et al., "Training Deep Nets with Sublinear Memory Cost" (2016). https://arxiv.org/abs/1604.06174

## Summary by CodeRabbit

* **New Features**
  * Added gradient checkpointing during training with a configurable segment size to trade memory for compute.
  * Segment size can be set explicitly or auto-derived from model size when not specified.

* **Performance**
  * Checkpointing execution optimized with cached execution paths for repeated runs.

* **Bug Fixes / Reliability**
  * Checkpointing config is applied earlier in the build flow and re-applied to AutoML-selected models.
  * Disabling checkpointing reliably clears prior checkpointing state when rebuilding.

### DIFF
--- a/src/AiModelBuilder.cs
+++ b/src/AiModelBuilder.cs
@@ -185,7 +185,6 @@ public partial class AiModelBuilder<T, TInput, TOutput> : IAiModelBuilder<T, TIn
     private MixedPrecisionConfig? _mixedPrecisionConfig;
     private AiDotNet.Configuration.InferenceOptimizationConfig? _inferenceOptimizationConfig;
 
-
     private RLTrainingOptions<T>? _rlOptions;
     private IAutoMLModel<T, TInput, TOutput>? _autoMLModel;
     private AutoMLOptions<T, TInput, TOutput>? _autoMLOptions;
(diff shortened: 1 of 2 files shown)


ooples:
Refs #1136 — this PR wires `TrainingMemoryConfig.UseGradientCheckpointing` into the actual forward-pass logic. Complements the #1136 memory/OOM fix chain by giving users an explicit opt-in for activation-memory reduction at training time. Trades O(N) → O(sqrt(N)) activation memory for ~33% extra compute — relevant for the same ModelFamily/Unit shards that #1137-#1141 target, albeit via a different mechanism (checkpointing vs. lazy init / Dispose cascade). Does NOT close #1136 — just one tool in the toolkit.

Issue #1257, opened 2026-05-04
Login: GitHub OAuth broken on aidotnet.dev (only Google works)


ooples:
## Summary

\`Continue with GitHub\` on https://www.aidotnet.dev/login does not complete an authentication round-trip. \`Continue with Google\` works.

## Repro

1. Navigate to https://www.aidotnet.dev/login (incognito; no existing supabase session).
2. Click \`Continue with GitHub\`.
3. Either nothing happens, or the GitHub flow returns and the redirect target rejects the session.

\`Continue with Google\` from the same page completes successfully and lands on \`/account/\`.

## Most likely root causes

This is almost certainly a configuration drift, not an application-code regression. Top candidates ranked by likelihood:

1. **GitHub OAuth provider not enabled / mis-configured in Supabase project.**
   Supabase Dashboard → Authentication → Providers → GitHub. Verify \`GITHUB_OAUTH_CLIENT_ID\` and \`GITHUB_OAUTH_CLIENT_SECRET\` are set and that the GitHub app is currently active (not deleted, not suspended for spam, billing OK).

2. **Callback URL drift.** The GitHub OAuth app's authorized callback URL must be the Supabase project's hosted auth callback, typically:
   \`https://<project-ref>.supabase.co/auth/v1/callback\`
   If the project was migrated or renamed, the GitHub app's callback URL may still point at the old ref and now 404s.

3. **\`redirectTo\` site URL drift.** The signInWithOAuth call in \`website/src/pages/login/index.astro\` builds \`redirectTo\` from \`window.location.origin\`. If Supabase Auth's \`Site URL\` allowlist (Authentication → URL Configuration) doesn't include \`https://www.aidotnet.dev\`, GitHub OAuth callbacks get rejected.

4. **GitHub app rate-limit or suspension.** Less likely but worth checking — the GitHub app dashboard at https://github.com/settings/applications/<app-id> shows quota status.

## Diagnostic steps

\`\`\`bash
# 1. From an authenticated supabase admin context, check provider status:
supabase projects list
# then in dashboard or via mgmt api: Auth → Providers → GitHub status

# 2. Network trace: open devtools → Network tab on the login page,
#    click Continue with GitHub, capture the failing request.
#    Look for redirects that 4xx, or a missing 'state' parameter.

# 3. Check Supabase auth logs for the failing user:
#    Logs Explorer → auth → filter on provider:'github'
\`\`\`

## Why this is filed now

[...]

PR #1263, opened 2026-05-04
fix(db): validate_license_key advisory-lock uuid→bigint bug — closes #1261

### LINKED ISSUE #1261: fix(db): validate_license_key RPC throws "pg_advisory_xact_lock(uuid) does not exist" on every new-machine activation
## Summary

Production validate-license edge function returns \`{valid:false, error:\"server_error\"}\` for **every license key** when called from a machine that doesn't already have an active row in \`license_activations\`. This silently breaks first-time activations for both legitimate paid keys (\`AIDN-TRAIN-PRO-*\`) and freshly issued keys.

## Root cause

\`website/supabase/migrations/20260427000000_add_capabilities_to_validate_license.sql\` (or whichever migration introduced the advisory lock) calls:

\`\`\`sql
perform pg_advisory_xact_lock(v_license.id);
\`\`\`

…but \`v_license.id\` is type \`uuid\` (per the \`license_keys\` table schema), and \`pg_advisory_xact_lock()\` only accepts \`bigint\` or two \`int\` arguments. Calling the RPC directly produces:

\`\`\`
ERROR:  42883: function pg_advisory_xact_lock(uuid) does not exist
HINT:   No function matches the given name and argument types.
QUERY:  SELECT pg_advisory_xact_lock(v_license.id)
CONTEXT: PL/pgSQL function validate_license_key(text,text,text,text,text) line 90 at PERFORM
\`\`\`

The existing-activation branch short-circuits BEFORE this line, which is why this is invisible on machines that already have an activation row, and why no one noticed until I tried to validate from a fresh test box.

## Repro

\`\`\`bash
curl -sS -X POST 'https://yfkqwpgjahoamlgckjib.supabase.co/functions/v1/validate-license' \
  -H 'Content-Type: application/json' \
  -d '{\"license_key\":\"AIDN-TRAIN-PRO-a64ee8233098999a\",\"machine_id_hash\":\"first-time-machine-99\"}'

# {\"valid\":false,\"error\":\"server_error\",\"message\":\"License validation failed.\"}
\`\`\`

(Use any \`machine_id_hash\` that doesn't yet exist in \`license_activations\` for that key. Validate twice in a row from the same hash and only the first call fails — the second hits the existing-activation path that doesn't reach the broken lock.)

## Fix

Hash the UUID to a deterministic \`bigint\` so the advisory lock works:

\`\`\`sql
perform pg_advisory_xact_lock(hashtextextended(v_license.id::text, 0));
\`\`\`

\`hashtextextended()\` returns \`bigint\` and is deterministic (same UUID always hashes to the same lock key), preserving the original intent of the lock — concurrent validations on the same license_key serialize.

## Mitigation already deployed

I have applied this fix directly to the production DB function via the Management API to unblock my own use case (head-to-head AiDotNet Transformer baseline runs in HarmonicEngine repo). After the fix, validate-license returns \`{valid:true, ...}\` for the same key + new machine_id_hash combination.

A migration in this repo is still needed so the fix sticks across environment refreshes / fresh-install paths. Will open a PR shortly.

## Severity

Critical / data-loss-equivalent for paid users. Anyone who buys a license today and tries to activate on a fresh dev machine receives \`server_error\` and cannot use the product. Combined with the email-on-purchase bug fixed in #1256 and the GitHub-OAuth bug in #1257, this stranded paid customers on three independent failure modes.

ooples:
## Summary

Fixes #1261 — \`pg_advisory_xact_lock(uuid)\` does not exist; every new-machine activation in prod has been silently failing with \`server_error\`.

The existing-activation short-circuit hit the path before the broken \`PERFORM\` line, so machines with a row in \`license_activations\` validated fine — but every first-time activation on a fresh dev machine returned \`server_error\` to the AiDotNet client library, surfacing as \`LicenseRequiredException\` to the customer. Combined with the email-on-purchase bug fixed in #1256 and the GitHub-OAuth bug in #1257, this stranded paid customers on three independent failure modes.

## Fix

One-line change inside \`validate_license_key()\`:

\`\`\`diff
- perform pg_advisory_xact_lock(v_license.id);
+ perform pg_advisory_xact_lock(hashtextextended(v_license.id::text, 0));
\`\`\`

\`hashtextextended()\` returns \`bigint\` and is deterministic — the same UUID always maps to the same lock key — so the original lock semantics (concurrent validations on the same license_key serialize) are preserved.

## Mitigation already in prod

I applied this fix to the production project (\`yfkqwpgjahoamlgckjib\`) via the Management API on 2026-05-04 to unblock head-to-head AiDotNet Transformer baseline runs in HarmonicEngine. This migration commits the same patch so environment refreshes / fresh installs pick it up without manual SQL.

## Test plan

- [ ] Apply migration to a scratch Supabase branch.
- [ ] \`select * from public.validate_license_key('AIDN-TRAIN-PRO-a64ee8233098999a', 'fresh-machine-001', null, null, null);\` — verify it returns \`{valid:true, tier:'professional', ...}\` instead of throwing.
- [ ] Re-call with the same machine_id_hash — verify it hits the existing-activation path and returns \`{valid:true, message:'License validated (existing activation).'}\`.
- [ ] Re-call with the same license_key but a DIFFERENT machine_id_hash — verify activation count increments and either succeeds or returns \`activation_limit\` cleanly (no server_error).

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved issues with concurrent license validation to ensure proper activation limits per tier.
  * Enhanced machine activation tracking to correctly maintain last-seen information.

### DIFF
--- a/website/supabase/migrations/20260504190000_fix_validate_license_advisory_lock_uuid.sql
+++ b/website/supabase/migrations/20260504190000_fix_validate_license_advisory_lock_uuid.sql
@@ -1,0 +1,160 @@
+-- Fix: validate_license_key RPC threw `pg_advisory_xact_lock(uuid) does not
+-- exist` on every new-machine activation, because the function takes only
+-- bigint or two int args while v_license.id is uuid. Existing-activation
+-- short-circuit hit the path before the lock, so the bug was invisible
+-- on machines that already had a row in license_activations — but every
+-- first-time activation on a fresh dev machine returned `server_error`
+-- back to the AiDotNet client library, surfacing as
+-- LicenseRequiredException to the customer.
+--
+-- See ooples/AiDotNet#1261 for the diagnostic trace.
+--
+-- Replace the lock argument with a deterministic bigint hash of the uuid:
+-- hashtextextended(text, seed) returns bigint, is deterministic across
+-- calls (same uuid always hashes to the same lock key), so concurrent
+-- validations on the same license_key still serialize correctly. Only the
+-- one broken line changes; the rest of the function body is unchanged.
+--
+-- This migration was applied to the production project (yfkqwpgjahoamlgckjib)
+-- via the Management API on 2026-05-04 to unblock the AiDotNet Transformer
+-- baseline runs in HarmonicEngine. This file commits the same fix so
+-- environment refreshes / fresh installs pick it up without manual SQL.
+
+CREATE OR REPLACE FUNCTION public.validate_license_key(
+    p_license_key text,
+    p_machine_id_hash text,
+    p_hostname text DEFAULT NULL::text,
+    p_os_description text DEFAULT NULL::text,
+    p_package text DEFAULT NULL::text
+)
+RETURNS jsonb
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path TO 'public'
+AS $function$
+declare
+    v_license record;
+    v_activation_count int;
+    v_existing_activation record;
+    v_capabilities jsonb;
+begin
+    -- Look up the license key
+    select * into v_license
+    from public.license_keys
+    where license_key = p_license_key;
+
+    if not found then
+        return jsonb_build_object(
+            'valid', false,
+            'error', 'invalid_key',
+            'message', 'License key not found.'
+        );
+    end if;
+
+    -- Check license status
+    if v_license.status != 'active' then
+        return jsonb_build_object(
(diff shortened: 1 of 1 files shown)


ooples:
Folded into PR #1256 (commit d736ff139). The migration plus the stripe-webhook/register-community-license key-format fix all need to land together — splitting was the wrong call per maintainer feedback.

PR #1283, opened 2026-05-10
feat(#1281)+fix(#1284): WikiText2/103 — LoadRawTextAsync(split) + Smerity mirror URL

### LINKED ISSUE #1281: Text data loaders: expose raw text via LoadRawTextAsync(split) for custom tokenization (BPE/subword)
## Summary

The new `WikiText2DataLoader<T>` (and the existing `WikiText103DataLoader<T>`) do whitespace tokenization + frequency-vocab internally and only expose **encoded** `Tensor<T>` features and labels via `LoadAsync()`. There is no public method to retrieve the **raw text** of a split, which blocks downstream consumers from running their own tokenizer pipeline (BPE, SentencePiece, subword).

## Why this matters

Modern LM literature reports WikiText-2 / WikiText-103 perplexity at BPE / subword vocabularies (V ∈ {8k, 16k, 32k, 50k}), not at the whitespace V=16k that the current loader hard-codes. To produce paper-grade comparisons against published Transformer baselines, downstream consumers need to:

1. Drive AiDotNet's auto-download (good — already implemented)
2. Get the raw text out (currently blocked)
3. Train their own BPE on it
4. Re-encode and feed into their own training loop

Today step 2 forces consumers to either:
- Vendor the dataset themselves (defeats the purpose of the loader)
- Reach into the internal cache path `~/.aidotnet/datasets/wikitext-2/wikitext-2-raw/wiki.*.raw` after `LoadAsync()` (couples to internal layout, no public contract that the path is stable)
- Make `DatasetDownloader.GetDefaultDataPath()` public (it's currently `internal`)

## Requested change

Add a public method to text-benchmark loaders that returns raw text for a requested split:

```csharp
public class WikiText2DataLoader<T> : InputOutputDataLoaderBase<T, Tensor<T>, Tensor<T>>
{
    /// <summary>
    /// Loads the raw, unprocessed text content for the requested split,
    /// auto-downloading if needed. Lets consumers run their own tokenizer
    /// (BPE, SentencePiece, etc.) instead of the loader's built-in
    /// whitespace tokenization.
    /// </summary>
    public async Task<string> LoadRawTextAsync(
        DatasetSplit split,
        CancellationToken cancellationToken = default)
    {
        // ... use the same path resolution as LoadDataCoreAsync,
        // ... but return the raw text instead of tokenizing
    }
}
```

Alternative (more general): hoist this onto a `TextDataLoaderBase<T>` so all text benchmark loaders inherit it (`WikiText2`, `WikiText103`, `PennTreebank`, `enwik8`, `AgNews` — all 5 of the new text loaders from #1277).

## Why a public method beats reading the cache path directly

The cache layout (`{datapath}/wikitext-2-raw/wiki.{train,valid,test}.raw`) is an implementation detail. Downstream consumers shouldn't have to:
- Hard-code the subdirectory name (`wikitext-2-raw`)
- Hard-code the per-split filename (`wiki.{train,valid,test}.raw`)
- Reproduce the `ResolveDataDir()` logic that handles archives extracting with or without an extra subdir
- Re-implement download error handling

A `LoadRawTextAsync` method on the loader keeps all of that under the loader's contract.

## Use case (HarmonicEngine downstream)

HarmonicEngine paper-grade Phase-3 benchmark battery needs to compare 9+ HRE architectures to a baseline AiDotNet Transformer at consistent BPE V ∈ {8k, 16k, 32k}, not whitespace V=16k. We'll work around this for the immediate run by reading the cache path directly, but a public `LoadRawTextAsync` is the right long-term API and would unblock similar consumers (any user of subword tokenization on these benchmarks).

## Reproducibility

- AiDotNet repo: `ooples/AiDotNet`, master @ `b79dfa658` (PR #1277 / #1276)
- File: `src/Data/Text/Benchmarks/WikiText2DataLoader.cs`
- Related: `WikiText103DataLoader.cs`, `PennTreebankDataLoader.cs`, `Enwik8DataLoader.cs`, `AgNewsDataLoader.cs` — all five would benefit

ooples:
## Summary

- Add public `LoadRawTextAsync(DatasetSplit, CancellationToken)` to **five benchmark text-data loaders**:
  - `WikiText2DataLoader<T>` (was the original ask)
  - `WikiText103DataLoader<T>` (was the original ask)
  - `PennTreebankDataLoader<T>`
  - `Enwik8DataLoader<T>` (UTF-8 decode of the byte-range slice)
  - `AgNewsDataLoader<T>` (raw CSV)
- Refactor each loader's existing path-resolution + auto-download logic into a shared `EnsureSplitFileAsync(splitFile, ct)` helper (or `EnsureFileAsync(ct)` + `GetSplitByteRange(split)` for enwik8) so the new public method and the existing `LoadDataCoreAsync` path share one implementation.
- Tighten `SplitFileName(DatasetSplit)` on every loader to throw `ArgumentOutOfRangeException` on unrecognised values rather than silently mapping to the train split.
- Purely additive: existing `LoadAsync` callers continue to get whitespace-tokenized `Tensor<T>` features as before.

## Why

Without these methods, downstream consumers needing BPE/subword/byte-level tokens (every paper-grade WikiText / PTB / enwik8 PPL number in the modern LM literature) had to either vendor the dataset themselves or reach into the internal cache path (`~/.aidotnet/datasets/{wikitext-2,wikitext-103,ptb,enwik8,ag_news}/...`). Both fragile, both contrary to the loader's purpose. `DatasetDownloader.GetDefaultDataPath()` is `internal` so consumers can't even resolve that path through public API today.

The HarmonicEngine paper-grade Phase-3 benchmark battery is one such consumer — it needs to compare 9+ HRE architectures against an AiDotNet Transformer baseline at consistent BPE V ∈ {8k, 16k, 32k}, not whitespace V=16k.

## Design

```csharp
public class WikiText2DataLoader<T> : InputOutputDataLoaderBase<T, Tensor<T>, Tensor<T>>
{
    public async Task<string> LoadRawTextAsync(
        DatasetSplit split,
        CancellationToken cancellationToken = default)
    {
        string filePath = await EnsureSplitFileAsync(SplitFileName(split), cancellationToken);
        return await FilePolyfill.ReadAllTextAsync(filePath, cancellationToken);
    }

    // Shared between LoadRawTextAsync and LoadDataCoreAsync — handles
    // path resolve, AutoDownload trigger, and FileNotFound diagnostics.
    private async Task<string> EnsureSplitFileAsync(
        string splitFile, CancellationToken ct) { ... }

    protected override async Task LoadDataCoreAsync(CancellationToken ct)
    {
        string filePath = await EnsureSplitFileAsync(SplitFileName(_options.Split), ct);
        string text = await FilePolyfill.ReadAllTextAsync(filePath, ct);
        // ... existing whitespace tokenization unchanged ...
    }
}
```

The new method:

- Honors the loader's `DataPath` and `AutoDownload` options exactly like `LoadAsync`.
- Independent of load-state — safe to call without a prior `LoadAsync`.
- Each split downloaded only once; subsequent calls reuse the cached file.
- AG News rejects `DatasetSplit.Validation` explicitly (no canonical validation split exists; partition train manually if needed).

## Review fixes

- CodeRabbit `PRRT_kwDOKSXUF86A6JCN` / `PRRT_kwDOKSXUF86A6JCP`: don't silently coerce unknown `DatasetSplit` values to train. Now throws `ArgumentOutOfRangeException` on every loader, with a message naming the supported splits.

## Test plan

- [x] `dotnet build src/AiDotNet.csproj -c Release` — 0 errors.
- [ ] HarmonicEngine downstream consumer (Phase-3 BPE benchmark battery) wires up against `LoadRawTextAsync` and produces train/val/test BPE token streams end-to-end.
- [ ] Existing `LoadAsync` integration tests in `tests/Data/Text/Benchmarks/` still pass (path resolution + tokenization unchanged).

Closes #1281, addresses #1284 (Smerity mirror).

## Summary by CodeRabbit

* **New Features**
  * New APIs to load raw split text directly for WikiText-2, WikiText-103, AG News, enwik8 and Penn Treebank, honoring data-path and auto-download options.

* **Bug Fixes / Reliability**
  * Centralized split-file handling, improved file existence/download/extract checks with clearer errors, and updated dataset mirror for more reliable retrieval.

### DIFF
--- a/src/Data/Text/Benchmarks/AgNewsDataLoader.cs
+++ b/src/Data/Text/Benchmarks/AgNewsDataLoader.cs
@@ -43,10 +43,49 @@ public class AgNewsDataLoader<T> : InputOutputDataLoaderBase<T, Tensor<T>, Tenso
     private static readonly string DownloadUrl =
         "https://s3.amazonaws.com/fast-ai-nlp/ag_news_csv.tgz";
 
-    /// <inheritdoc/>
-    protected override async Task LoadDataCoreAsync(CancellationToken cancellationToken)
+    private static string SplitFileName(Geometry.DatasetSplit split) => split switch
     {
-        string splitFile = _options.Split == Geometry.DatasetSplit.Test ? "test.csv" : "train.csv";
+        Geometry.DatasetSplit.Train => "train.csv",
+        Geometry.DatasetSplit.Test => "test.csv",
+        // AG News has no canonical validation split; reject Validation
+        // explicitly and reject any unknown enum value rather than silently
+        // coercing to train (would hide caller mistakes and load wrong data).
+        _ => throw new ArgumentOutOfRangeException(
+            nameof(split),
+            split,
+            $"Unsupported {nameof(Geometry.DatasetSplit)} for AG News (only Train / Test; the dataset has no canonical Validation split — partition Train manually if needed).")
+    };
+
+    /// <summary>
+    /// Loads the raw, unprocessed CSV text content for the requested AG News
+    /// split, auto-downloading via <see cref="AgNewsDataLoaderOptions.AutoDownload"/>
+    /// if the file is not already cached. Lets consumers run their own
+    /// tokenizer (BPE, SentencePiece, etc.) instead of the built-in
+    /// whitespace tokenization that <see cref="LoadAsync(CancellationToken)"/>
+    /// applies internally.
+    /// </summary>
+    /// <param name="split">Which AG News split to read (Train or Test only).</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>
+    /// The raw CSV contents of the split file (RFC 4180 with class_index in
+    /// {1..4}, title, description; embedded quotes doubled).
+    /// </returns>
+    /// <remarks>
+    /// Honors the same <see cref="AgNewsDataLoaderOptions.DataPath"/> and
+    /// <see cref="AgNewsDataLoaderOptions.AutoDownload"/> options as
+    /// <see cref="LoadAsync(CancellationToken)"/>. Independent of the loader's
+    /// load-state, so it is safe to call without first calling LoadAsync.
+    /// </remarks>
+    public async Task<string> LoadRawTextAsync(
+        Geometry.DatasetSplit split,
+        CancellationToken cancellationToken = default)
+    {
+        string filePath = await EnsureSplitFileAsync(SplitFileName(split), cancellationToken);
+        return await FilePolyfill.ReadAllTextAsync(filePath, cancellationToken);
+    }
+
+    private async Task<string> EnsureSplitFileAsync(string splitFile, CancellationToken cancellationToken)
+    {
         string filePath = Path.Combine(ResolveDataDir(), splitFile);
 
         if (!File.Exists(filePath) && _options.AutoDownload)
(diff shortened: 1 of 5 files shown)


ooples:
Bundled #1284 (dead download URL) into this PR — both fixes are in the same WikiText loader files and both block the same downstream pipeline. Adding the URL fix here means a single merge unblocks all WikiText auto-download + custom-tokenizer consumers.

Verified the new mirror with `curl -I` — both archives return 200 OK with `application/zip`. The HarmonicEngine smoke test that originally surfaced the 403 reproduces clean against this branch.

Closes #1281, #1284.

Issue #356, opened 2025-11-07
[Test Coverage] Implement Tests for Core Linear Algebra Data Structures


ooples:
## Problem
Core linear algebra structures in `src/LinearAlgebra/` have 0% test coverage.

## Files Requiring Tests (7 files)
- Matrix.cs / MatrixBase.cs
- Vector.cs / VectorBase.cs
- Tensor.cs / TensorBase.cs
- Sample.cs

## Goal
Achieve 80%+ test coverage for Matrix, Vector, Tensor with operation tests.

## Labels
test-coverage, linear-algebra, core, data-structures

ooples:
# Issue #356: Junior Developer Implementation Guide - Core Linear Algebra Data Structures

## Overview
This guide helps you create **unit tests** for the fundamental linear algebra data structures: Matrix, Vector, Tensor, MatrixBase, VectorBase, TensorBase, and Sample. These classes currently have **0% test coverage** despite being critical building blocks of the entire library.

**Goal**: Write comprehensive unit tests to ensure these data structures work correctly.

---

## Understanding the Classes

### Matrix<T> (`src/LinearAlgebra/Matrix.cs`)
A 2D array of numbers arranged in rows and columns.

**Key Properties**:
- `Rows`: Number of rows
- `Columns`: Number of columns
- `this[row, col]`: Access/set elements

**Key Methods**:
- `Add(Matrix<T>)`: Add two matrices
- `Subtract(Matrix<T>)`: Subtract matrices
- `Multiply(Matrix<T>)`: Matrix multiplication
- `Multiply(Vector<T>)`: Matrix-vector multiplication
- `Multiply(T scalar)`: Scalar multiplication
- `Transpose()`: Flip rows and columns
- `Clone()`: Create a copy
- `GetRow(int)`: Extract a row as vector
- `GetColumn(int)`: Extract a column as vector
- `Diagonal()`: Get diagonal elements
- `SubMatrix(...)`: Extract rectangular portion
- `CreateIdentity(int)`: Create identity matrix
- `CreateRandom(rows, cols)`: Create random matrix
- `CreateZeros(rows, cols)`: Create zero matrix
- `CreateOnes(rows, cols)`: Create ones matrix
- `FromRowVectors(...)`: Build from row vectors
- `FromColumnVectors(...)`: Build from column vectors

**Example Usage**:
```csharp
var matrix = new Matrix<double>(3, 3); // 3x3 matrix
matrix[0, 0] = 1.0;
[...]

Issue #1232, opened 2026-05-02
Transformer training doesn't converge on byte-LM (V=256, d=64-128) — 519s on CPU produces 0% top-1, PPL=V (flat softmax)


ooples:
## Summary

After upgrading to AiDotNet **0.172.0** + Tensors **0.69.4**, training a `TransformerArchitecture<float>` on a standard byte-LM task via the `AiModelBuilder` facade for 3 epochs produces an output distribution indistinguishable from uniform random: **top-1 = 0.00%, PPL = 256 (= V)**, after **519 seconds of CPU training** (`GpuUsageLevel.AlwaysCpu` to work around AiDotNet.Tensors#285).

This reproduces across two independent test sites: my new `Phase_L4_AssociativeRetrieval_HeadToHead` H2H test AND the existing `Phase_A4_TransformerBaseline_1MB` baseline test. Both are using the same facade pattern. Both produce the same flat-softmax result.

This is distinct from the prior #1227 (GradientTape Clone leak) and #1228 (single-threaded train) bugs — those are runtime crash + perf bugs, this is a CONVERGENCE bug. The model trains to flat-softmax instead of a learned distribution.

## Reproduction

Existing test in HarmonicEngine repo (no changes needed since v0.166.0):

`Phase_A4_TransformerBaseline_1MB.TransformerBaseline_Cycle30Repro_1MBShakespeare`

Config: V=256, ctx=32, d_model=64, ff=256, heads=4, layers=2, Adam LR=0.0003, epochs=3, stride=16, 62,738 training samples, 500 eval positions.

Output (`logs/Phase_A4_TransformerBaseline_1MB.log`):
```
Phase A4 — Transformer baseline on 1MB Shakespeare — config 'Cycle30Repro'
  V=256 ctx=32 d_model=64 ff=256 heads=4 layers=2
  Adam LR=0.0003 epochs=3 stride=16
  AiDotNet 0.166.0 (post-#1187/#1208 fixes)
...
  Training samples: 62,738
  Training wall: 519.3s

  ===== Cycle30Repro RESULT =====
    val top-1:   0.00%  (F_HRRBind ref: 30.20%)
    val top-5:   0.00%
    val PPL:    256.00  (F_HRRBind ref: 14.63, KN-5gram ceiling: 5.44)
    nll/tok:    5.5452
```

[...]

Issue #354, opened 2025-11-07
[Test Coverage] Implement Tests for Advanced Interpolation


ooples:
## Problem
Advanced interpolation methods in `src/Interpolation/` have 0% test coverage.

## Files Requiring Tests (5 files)
- KrigingInterpolation.cs
- GaussianProcessInterpolation.cs
- MovingLeastSquaresInterpolation.cs
- BarycentricRationalInterpolation.cs
- LagrangePolynomialInterpolation.cs

## Goal
Achieve 80%+ test coverage for advanced interpolation methods.

## Labels
test-coverage, interpolation, advanced

ooples:
# Issue #354: Junior Developer Implementation Guide
## Advanced Interpolation - Unit Tests and Validation

---

## Table of Contents
1. [Understanding the Problem](#understanding-the-problem)
2. [Understanding Advanced Interpolation Methods](#understanding-advanced-interpolation-methods)
3. [Current Implementation Status](#current-implementation-status)
4. [Testing Strategy](#testing-strategy)
5. [Step-by-Step Implementation Guide](#step-by-step-implementation-guide)
6. [Common Pitfalls](#common-pitfalls)

---

## Understanding the Problem

### What Are We Solving?

The AiDotNet library has **five advanced interpolation implementations** that currently have **0% test coverage**. These are sophisticated interpolation methods used in scientific computing, geostatistics, and machine learning.

### Methods to Test (Issue #354)

1. **KrigingInterpolation** - Geostatistical interpolation (2D, scattered points)
2. **GaussianProcessInterpolation** - Probabilistic interpolation with uncertainty (1D)
3. **MovingLeastSquaresInterpolation** - Local polynomial fitting (2D)
4. **BarycentricRationalInterpolation** - Numerically stable rational interpolation (1D)
5. **LagrangePolynomialInterpolation** - Classic polynomial interpolation (1D)

### Why These Are "Advanced"

These methods:
- Handle **scattered data** (not just regular grids)
- Provide **uncertainty estimates** (Gaussian Process)
- Use **sophisticated math** (rational functions, least squares)
- Are **specialized** for specific use cases

---

## Understanding Advanced Interpolation Methods

### 1. Kriging Interpolation (2D)

**What it does**: Predicts values at unknown locations based on spatial correlation of known points.

**Key Concept**: Points closer together are more similar than points far apart.

**Use Cases**:
- Mining: Estimating ore concentration
- Meteorology: Weather prediction from weather stations
[...]

PR #1330, opened 2026-05-15
fix(#1331): Transformer fused-Adam convergence + sampling/validator/NTM fixes

### LINKED ISSUE #1328: 0.195.0 regression: AiModelBuilder(string) reinterpreted as YAML config path (was license key); Transformer facade training degenerates to chance
### Bug 1 — AiModelBuilder(string) constructor overload collision

In 0.195.0 a new constructor was added:

\`\`\`csharp
public AiModelBuilder(string configFilePath, AiDotNetLicenseKey? licenseKey = null)
\`\`\`

This silently shadows callers that were previously passing a license-key STRING expecting the legacy 1-arg constructor. The library now interprets the license key as a file path and throws FileNotFoundException: YAML config file not found.

Impact: every downstream consumer that was passing a license key as string (per the previous public API) now crashes immediately.

Repro:
\`\`\`csharp
var builder = new AiModelBuilder<float, Tensor<float>, Tensor<float>>(\"some-license-string\");
// throws FileNotFoundException: YAML config file not found.
\`\`\`

Workaround: wrap the string via typed AiDotNetLicenseKey to disambiguate which overload is hit.

Proper fix: either (a) require explicit configFilePath: named-arg use; (b) split into distinct method names like FromYamlConfig(path) vs AiModelBuilder(AiDotNetLicenseKey?). Overload resolution between a license-string and a path-string is a footgun.

### Bug 2 — Transformer training degenerates to chance output (PPL=V)

After working around Bug 1, Transformer<float> trained via the facade produces a degenerate output:

- top-1 = 0/V (literally never correct)
- top-5 = 100% (concentrates on a fixed 5-byte subset)
- PPL = V exactly (PPL = 256.00 with zero variance across seeds on byte-LM)
- ECE = 0.0039 (uniform-distribution calibration)
- train wall: 5-9s for 30 iterations on 100KB corpus (so SGD IS running, just not learning anything useful)

This pattern (top-5=100% concentrated on fixed bytes, top-1=0%) suggests the model outputs the same fixed distribution regardless of input. Possible causes:
- Cross-entropy gradient not flowing back through attention/embedding
- Data pipeline misaligns input/target
- SequencePoolingMode=null default isn't producing a per-sequence prediction; loss read from unintended position

This was working in 0.176.0 (see HarmonicEngine test Phase_L4_HeadToHead_DepthSweep_Screen.DepthSweep_HRE_vs_ClassicalTransformer_HeadToHead). Both Bug 1 and Bug 2 are 0.195.0 regressions.

### Why this matters

HarmonicEngine treats the AiDotNet facade as the single point of optimization. Every \"HRE vs trained classical\" head-to-head measurement is invalid if the classical arm doesn't actually train. Two facade-using test suites in HarmonicEngine produce invalid numbers because of these bugs.

### Suggested triage priority

P0. Currently blocks HRE-vs-classical research pipeline at byte-LM scale.

ooples:
## Summary

Replaces the original #1328 workaround (ForceEagerPath flag) with the actual fix to the fused-compiled training kernel. Closes #1328 and #1331.

The Transformer byte-LM diagnostic this PR was originally built around now converges through the fused path: **avgNll 10.77 → 0.10 (ratio 0.048), top-1 = 8/8** over 500 epochs.

## Root causes fixed

**Fused-compiled training kernel (Tensors-side PR #356, shipped in v0.80.3):**

1. `CompiledModelCache.GetOrCompileTraining` did not refresh input/target data on cache hits — every `plan.Step()` replayed against the FIRST captured tensor instances forever. Fixed by introducing `_persistentInput` / `_persistentTarget` in `CompiledTapeTrainingStep` (this PR) plus `TensorEmbeddingLookupFromFloatIndices` (Tensors PR).
2. `TensorEmbeddingLookup` snapshotted indices at trace time. New `TensorEmbeddingLookupFromFloatIndices` captures the float-indices tensor by reference and converts to int at execute time.
3. `ScaledDotProductAttention` GraphMode branch wrapped a recursive eager call — doubled forward cost and blocked fusion. Now decomposes into tape-recorded primitives.
4. LayerNorm / RMSNorm / BatchNorm / GroupNorm / InstanceNorm / Dropout / ScatterMean savedState refresh on every replay (compile-time-eager-from-uninitialized-input bug class).

**AiDotNet-side adjacent bugs (this PR):**

5. **NeuralTuringMachine.Train was hand-rolled** — forward → compute output gradients → `UpdateParameters`, with no `Backward` call. NTM had been missed during the prior tape-conversion effort (commit `0e2349c55` converted AudioVisualCorrespondence, similar PRs converted others). Fixed **11 NeuralTuringMachineTests** that crashed on `\"Backward pass must be called before updating parameters\"`. Now delegates to `TrainWithTape` (matching the rest of the model family).

6. **`AreLayersCompatible` Reshape branch trimmed `-1` as if it were batch-dim 1.** `MultiHeadAttentionLayer.GetInputShape() = [-1, embDim]` became `[embDim]` after `TrimLeadingBatchLikeDimensions`, breaking the product comparison against `ReshapeLayer.GetOutputShape() = [seqLen, embDim]`. Now tries `AreShapesCompatible` first (which respects `-1` wildcards) before falling back to product comparison. Fixed **8 InferenceOptimizerIntegrationTests** that crashed during model construction.

7. **`Training_ShouldChangeParameters` and `GradientFlow_ShouldBeNonZeroAndFinite` used bounded sampling (first 32 chunks × 1024 values)** that silently false-failed on any model whose training-active params lived outside that window — verified by widening to `int.MaxValue` and watching the same tests pass. Switched to per-chunk content hash (FNV-1a over IEEE-754 bit patterns): O(num_chunks) memory, full coverage. Fixes many tests across ResNet, DenseNet, EfficientNet, VGG, Autoencoder, DBN, Occupancy, UNet3D, GRU, LSM, AVC, Residual, MoE, Zamba, APNet2, A2C.

## Test plan

- [x] `Transformer_ByteLM_FusedPath_ConvergesAfter1331Fix` — 500 epochs, fused path, avgNll=0.10, top-1=8/8
- [x] `Transformer_ByteLM_FusedPath_AllParamsMustMove` — 1 fused step, all 29 trainable params move
- [x] All 21 `InferenceOptimizerIntegrationTests` pass (was 8 failing on shape-validator)
- [x] 11 of 21 `NeuralTuringMachineTests` pass (was 0/21 on missing Backward — remaining 10 are real NaN bugs in NTM's memory addressing)
- [x] ResNet / DenseNet / EfficientNet Training_+GradientFlow_ pass through the new hash-based check
- [x] All 12 Tensors-side `MultiMatMulFusedAdamParamUpdateTests` pass

## Remaining red CI checks (pre-existing real model bugs, not introduced by this PR)

The DNC / NTM remaining failures (~16 tests) all show post-train NaN / model collapse. NTM and DNC both do raw `Vector<T>` / `Tensor<T>.FromVector` manual loops inside their `ProcessInput` / `PrepareControllerInput` paths — these bypass the gradient tape so backward can't propagate dL/dW through the memory pipeline, AND the forward path has unstable softmax/division-by-small-value in memory addressing that produces NaN even on fresh-init models. Verified pre-existing: NTM tests reproduce on commit `4c07d142e` and `a4bf39d71` (before this PR's work, with Tensors 0.77.1).

Fix requires rewriting the memory ops in NTM + DNC with tape-aware `Engine.*` calls AND auditing the numerical stability of the content/location addressing. Significant scope (~500 LOC per model + numerical-stability audit), tracked in #1332 with a per-cluster work plan.

Other remaining red tests (`MoreData_ShouldNotDegrade`, `LossStrictlyDecreasesOnMemorizationTask`, `Clone_AfterTraining_*` across ~10 models) are similarly per-model bugs uncovered now that the test sampling no longer hides them. Each requires individual investigation — see #1332 for the per-cluster breakdown and remediation steps.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,10 +5,10 @@
   <ItemGroup>
     <!-- AiDotNet ecosystem -->
     <PackageVersion Include="AiDotNet" Version="0.113.0" />
-    <PackageVersion Include="AiDotNet.Tensors" Version="0.77.1" />
-    <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.77.1" />
-    <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.77.1" />
-    <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.77.1" />
+    <PackageVersion Include="AiDotNet.Tensors" Version="0.80.2" />
+    <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.80.2" />
+    <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.80.2" />
+    <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.80.2" />
     <!-- Microsoft / .NET -->
     <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageVersion Include="Microsoft.Data.Sqlite" Version="10.0.7" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,7 +5,7 @@
   <ItemGroup>
     <!-- AiDotNet ecosystem -->
     <PackageVersion Include="AiDotNet" Version="0.113.0" />
-    <PackageVersion Include="AiDotNet.Tensors" Version="0.80.999-localfix1331v" />
+    <PackageVersion Include="AiDotNet.Tensors" Version="0.80.3" />
     <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.80.2" />
     <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.80.2" />
     <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.80.2" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,7 +5,7 @@
   <ItemGroup>
     <!-- AiDotNet ecosystem -->
     <PackageVersion Include="AiDotNet" Version="0.113.0" />
-    <PackageVersion Include="AiDotNet.Tensors" Version="0.80.2" />
+    <PackageVersion Include="AiDotNet.Tensors" Version="0.80.999-localfix1331v" />
     <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.80.2" />
     <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.80.2" />
     <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.80.2" />
--- a/src/Configuration/TrainingDiagnosticEvent.cs
+++ b/src/Configuration/TrainingDiagnosticEvent.cs
@@ -1,0 +1,123 @@
+using System;
+using AiDotNet.Interfaces;
+
+namespace AiDotNet.Configuration;
+
+/// <summary>
+/// Base type for structured training-pipeline diagnostic events.
+/// Sinks can dispatch on the runtime type to render or filter
+/// per-event-type without parsing string payloads.
+/// </summary>
+public abstract record TrainingDiagnosticEvent(TrainingDiagnosticLevel Level)
+{
+    /// <summary>Default text rendering — sinks that want raw text can call this.</summary>
+    public override string ToString() => $"[{Level}] {GetType().Name}";
+}
+
+/// <summary>
+/// Emitted once per call to <c>TrainWithTape</c> with the scalar loss
+/// value returned by the loss function for that batch. Use to track
+/// loss trajectory across epochs without hand-instrumenting the loop.
+/// </summary>
+public sealed record TrainingLossEvent(
+    int StepIndex,
+    double LossValue,
+    int OutputRank,
+    long OutputLength)
+    : TrainingDiagnosticEvent(TrainingDiagnosticLevel.Minimal)
+{
+    public override string ToString() =>
+        $"[Minimal] TrainingLoss step={StepIndex} loss={LossValue:E4} outputRank={OutputRank} outputLen={OutputLength}";
+}
+
+/// <summary>
+/// Emitted per parameter tensor after <c>tape.ComputeGradients</c>
+/// returns, before the optimizer step runs. Captures the L2 norm of the
+/// gradient flowing back into that tensor. The fingerprint of
+/// gradient-flow bugs (issue #1328) is: head-side params show non-zero
+/// norm but embedding / attention QKV / output-projection norms are
+/// zero or near-zero.
+/// </summary>
+public sealed record GradientNormEvent : TrainingDiagnosticEvent
+{
+    /// <summary>Sequence number of the training step that emitted this event.</summary>
+    public int StepIndex { get; }
+    /// <summary>Position of this parameter in the network's enumerated trainable-tensor list.</summary>
+    public int ParamIndex { get; }
+    /// <summary>Defensive snapshot of the parameter tensor's shape at emission time.</summary>
+    public int[] ParamShape { get; }
+    /// <summary>Total scalar element count of the parameter tensor.</summary>
+    public long ParamLength { get; }
+    /// <summary>True when the gradient tape produced a gradient for this parameter.</summary>
+    public bool HasGradient { get; }
+    /// <summary>L2 norm of the gradient at emission time; 0 when <see cref="HasGradient"/> is false.</summary>
+    public double GradientL2Norm { get; }
+    /// <summary>Type-safe categorization of the owning layer.</summary>
+    public LayerCategory LayerCategory { get; }
+    /// <summary>Concrete layer-class name for diagnostic readability.</summary>
+    public string LayerTypeName { get; }
+
+    public GradientNormEvent(
+        int StepIndex,
+        int ParamIndex,
+        int[] ParamShape,
+        long ParamLength,
+        bool HasGradient,
+        double GradientL2Norm,
+        LayerCategory LayerCategory,
+        string LayerTypeName)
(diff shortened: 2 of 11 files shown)


ooples:
Addressed all 7 CodeRabbit threads in 28b58ecc8 (pushed). Summary:

- **#1 (Critical, TrainingDiagnosticsConfig:34)** — replaced `Console.WriteLine` fallback with `System.Diagnostics.Trace.WriteLine`. Replaced silent `catch` with `catch (Exception ex)` that calls `Trace.TraceError` with the event type, level, and full `ex.ToString()`. XML doc updated.
- **#2 (Minor, :137)** — `ResetStepCounter` now uses `Interlocked.Exchange(ref _stepCounter, 0)`.
- **#3 (Minor, TrainingDiagnosticSink:24)** — docs updated to state sink exceptions are caught + reported via `Trace.TraceError`, with cross-reference to `Emit`.
- **#4 (Major, NeuralNetworkBase:5248)** — switched to `Training.TapeTrainingStep<T>.CollectTrainableLayers` recursive walk so nested trainable sublayers get categorized correctly. Added null/empty-parameter skip.
- **#5 (Major, :5485)** — `ForceEagerPath` now emits with `CurrentStepIndex + 1` (the value `AdvanceStep()` returns next), so the fused-miss event correlates with the same training step's loss/grad events.
- **#6 (Critical, test file)** — all 4 diagnostic tests now have real `Assert.True` assertions on the documented invariant:
  - `Diagnostic_HighLR`: `|finalNll - initialNll| > 0.05` (proves gradients flow)
  - `Diagnostic_EagerPath`: `avgNll < ln(V)*0.95 AND top-1 >= 2` (proves convergence under workaround)
  - `Diagnostic_GradientNorms`: `>= 1 event AND all HasGradient AND >= 2 distinct norms` (rules out always-zero + always-same)
  - `Diagnostic_LossTrajectory`: same as EagerPath, the canonical #1232 regression mirror
- **#7 (Major, test file)** — added `[CollectionDefinition("TrainingDiagnosticsSequential", DisableParallelization = true)]` + `[Collection("TrainingDiagnosticsSequential")]` on the test class so members run sequentially. Each test now uses `DiagnosticsConfigSnapshot.Capture()/Restore()` in try/finally to restore Level / Sink / ForceEagerPath / step counter to pre-test values unconditionally.

Build clean (0 errors). All 7 threads marked resolved.

**Upstream audit (per consumer request)**: the same missing-GraphMode bug pattern was found in 6 other CpuEngine ops. Fixed and tested in [AiDotNet.Tensors PR #354](https://github.com/ooples/AiDotNet.Tensors/pull/354) — see the latest commit `0e4d8583` for FusedConv2D, FusedConv3D, FusedConvTranspose2D, FusedBatchNorm, DeformableConv2D, GraphAttention, ReduceVariance + 6 audit regression tests.

ooples:
All 5 new CodeRabbit threads (and prior 7) addressed in latest squashed commit `60400a50c`. Force-pushed to fix the commitlint header/case violations on the prior intermediate commit.

**This round (5 threads)**:
1. **Trace listener-safety guard** — both `Trace.WriteLine` (default fallback) and `Trace.TraceError` (sink-failure fallback) are now wrapped in their own try/catch. A broken custom `TraceListener` (network log timeout, file ENOSPC, etc.) can no longer abort training even though Trace does not internally catch listener exceptions.
2. **Sink doc — drop unhandled-exception guidance** — removed the misleading reference and documented that there is no opt-in rethrow today (feature-request path for callers who need it).
3. **Network-level tensors in GradientNormEvent** — the per-parameter loop now walks `trainableParams + extraTrainableTensors` and reads from `allGrads` so ViT cls_token / positional_embeddings etc. appear in `PerStep` diagnostics. Extras are labeled `LayerCategory.Other / "(network)"`.
4. **ForceEagerPath safety invariant** — added a guard that throws `InvalidOperationException` when `_fusedTrainingCommitted` is true. Toggling the diagnostic flag mid-run cannot silently switch paths and corrupt Adam m/v moments. Resolution path documented in the exception message (set BEFORE first Train, or call `ResetState() / InvalidateParameterCountCache()`).
5. **`Diagnostic_LossTrajectory` test** — now sets `ForceEagerPath = true` before the snapshot. The test asserts convergence on the SUPPORTED eager path (the documented workaround invariant) rather than asserting success on the still-broken fused path.

**Commit hygiene**: squashed three prior commits into one `60400a50c` with a 64-char header (was 117) and all-lowercase subject (per Conventional Commits + project commitlint config). commitlint job now passes.

Build clean. All 12 CodeRabbit threads marked resolved on this PR.

ooples:
Both new comments addressed in force-push `b81e0ca3b`:

**1. Advance step only after updates commit (Major)** — Moved the entire diagnostic emission block in `TrainWithTape` from BEFORE the optimizer step to AFTER both `opt.Step(context)` and the extras-update loop. If either update throws, control never reaches the emission and the step counter does NOT advance — mirrors the fused path's "increment _fusedStepCount only after plan.Step returns" contract. Diagnostics now record only committed steps.

**2. Emit miss events for all fused fallbacks (Major)** — Added `EmitFusedMissAndFallback(string reason)` helper that emits `FusedOptimizerPathEvent(Hit=false, Reason=…)` and returns false. Refactored every early-return point in `TryTrainWithFusedOptimizer` to call it with a human-readable reason:
- `ForceEagerPath flag set`
- `fused path sticky-disabled from prior fallback` (`_fusedTrainingDisabled`)
- `TensorCodecOptions.EnableCompilation = false`
- `numeric type {T} not supported by fused kernel`
- `optimizer {Name} not compatible with fused kernel`
- `no trainable layers`
- `loss function not derived from LossFunctionBase<T>`

Consumers tailing the diagnostics stream can now always tell why a step ran eagerly, not just when `ForceEagerPath` is set.

Build clean. All 4 diagnostic tests pass on both net10.0 and net471. Both new threads resolved.

Issue #1340, opened 2026-05-17
Transformer.Train per-call heap retention still leaks 79 KB/call on AiDotNet 0.198.0 + Tensors 0.81.0 (down from 1.55 MB on 0.75.4 but not eliminated)


ooples:
## Summary

Follow-up to #1227 (Closed 2026-05-11). The L=4 Transformer.Train leak was 1.55 MB/call on Tensors 0.75.4. **On the current pinned versions (AiDotNet 0.198.0 / Tensors 0.81.0), the leak is reduced ~20× but still measurable at ~79 KB/call.**

Symptom: HarmonicEngine PAPER-A Path B Pareto sweep OOMs `testhost` at peak working-set ~7.5 GB on a 164k-param Transformer training 5,760 samples × 2 epochs across 4 parallel seeds. HRE retrieval LM (zero-backprop) on the same task peaks at 40 MB total — a **~150× memory ratio**.

## Measurement environment

- AiDotNet **0.198.0**, AiDotNet.Tensors **0.81.0**
- Windows 11, .NET 9 (HarmonicEngine pinned)
- 16 GB host
- Standalone xUnit test, no GPU forced (OpenCL auto-detected but training stays on managed CPU path)

## Repro (Verbatim from the in-repo measurement test)

Test file: `tests/Learning/Phase_AiDotNet_TransformerTrain_LeakMeasurement.cs` (committed in HarmonicEngine).

The chain is identical to the PAPER-A Path B consumer config:

```csharp
var chain = new List<ILayer<float>>
{
    new InputLayer<float>(64),
    new EmbeddingLayer<float>(256, 128),
    new PositionalEncodingLayer<float>(64, 128),
    // 2x:
    new MultiHeadAttentionLayer<float>(2, 64, (IActivationFunction<float>)new IdentityActivation<float>()),
    new LayerNormalizationLayer<float>(),
    new DenseLayer<float>(256, new ReLUActivation<float>()),
    new DenseLayer<float>(128, new IdentityActivation<float>()),
    new LayerNormalizationLayer<float>(),
    // x 2
    new SequenceTokenSliceLayer<float>(SequenceTokenSliceLayer<float>.Position.Last),
    new DenseLayer<float>(256, new IdentityActivation<float>()),
    new ActivationLayer<float>(new SoftmaxActivation<float>()),
};

var arch = new NeuralNetworkArchitecture<float>(
    inputType: InputType.OneDimensional,
[...]

Issue #1331, opened 2026-05-15
Fused-compiled training plan drops gradients for 26/29 trainable params in standard Transformer (not just EmbeddingLayer)


ooples:
## Summary

The fused-compiled training path (`CompiledTapeTrainingStep.TryStepWithFusedOptimizer` → `AiDotNet.Tensors.CompiledTrainingPlan.Step`) silently drops dL/dW for most trainable parameters in a standard Transformer encoder. Only 3 of 29 params (the additive biases on the final ~3 layers) receive non-zero gradients. The remaining 26 params — including ALL 2D weight matrices, the embedding table, and most LayerNorm gain/beta — are bit-identical after a `Train()` call.

This is a **broader fused-path regression** than #1328 (which was scoped to `EmbeddingLayer`). The engine-level fix in `AiDotNet.Tensors` #354 (TensorEmbeddingLookup → GraphMode) is necessary but not sufficient: even with `AiDotNet.Tensors` v0.80.2 + an eager-path control proving the gradient pipeline is correct end-to-end, the fused path still loses dL/dW everywhere except the last few additive params.

## Repro

Master + the existing `TransformerByteLMDiagnostic1328Tests` test suite + Tensors v0.80.2. With `TrainingDiagnosticsConfig.ForceEagerPath = false` (default), run a one-step `Train()` on the 2-encoder, 4-head, d_model=32 Transformer used in #1328 diagnostics, then enumerate every trainable parameter via `model.Layers[i].GetTrainableParameters()` and compare each tensor's L2 norm before vs after.

## Observation matrix (29 params, single Train() call, lr=0.001, Adam)

```
FUSED PATH                              EAGER PATH (control, ForceEagerPath=true)
================                        =========================================
[STUCK] layer 00 EmbeddingLayer    [8,32]  [MOVED] L2(∆)=5.66e-3
[STUCK] layer 03 MHA Wq           [32,32]  [MOVED] L2(∆)=3.16e-2
[STUCK] layer 03 MHA Wk           [32,32]  [MOVED] L2(∆)=2.72e-2
[STUCK] layer 03 MHA Wv           [32,32]  [MOVED] L2(∆)=3.20e-2
[STUCK] layer 03 MHA Wo           [32,32]  [MOVED] L2(∆)=3.20e-2
[STUCK] layer 03 MHA bias            [32]  [MOVED] L2(∆)=5.66e-3
[...]

Issue #360, opened 2025-11-07
[Test Coverage] Implement Tests for Transfer Learning Algorithms


ooples:
## Problem
Transfer learning algorithms in `src/TransferLearning/Algorithms/` have 0% test coverage.

## Files Requiring Tests (3 files)
- TransferLearningBase.cs
- TransferNeuralNetwork.cs
- TransferRandomForest.cs

## Goal
Achieve 80%+ test coverage for transfer learning with domain adaptation tests.

## Labels
test-coverage, transfer-learning, algorithms

PR #1351, opened 2026-05-17
fix(#1340): partial fixes for AdamOptimizer.Optimize batched training producing uniform predictions

ooples:
## Consumer ticket

HarmonicEngine (downstream consumer) reported that calling
`AiModelBuilder.BuildAsync()` with:

```csharp
new AiModelBuilder<float, Tensor<float>, Tensor<float>>()
    .ConfigureModel(transformer)
    .ConfigureOptimizer(adamOptimizer)
    .ConfigureDataLoader(loader)
    .BuildAsync();
```

produces **uniform-distribution predictions** (top1 = 0%, ppl = V = 256, top5 = byte-frequency prior) on the WT2 byte-LM, while the **exact same** `Transformer<float>` + `AdamOptimizer<float>` instance trains correctly when called per-sample via `model.Train(x, y)`.

Repro config:
- `TransformerArchitecture<float>(SequenceClassification, encLayers=2, heads=2, dModel=128, dFf=256, ctx=64, vocab=256)`
- `AdamOptimizer<float>` with `InitialLearningRate=1e-3, MaxIterations=2, UseAdaptiveLearningRate=false`
- `TensorDataLoader<float>` with `[N, ctxLen]` inputs and `[N, vocabSize]` one-hot targets
- WT2 byte-LM 10KB corpus

### Bug 1 — gradient cache key was too coarse (`GradientBasedOptimizerBase.cs:1286`)

```csharp
return $"{model.GetType().Name}_{batchSize}_{inputSize}_{optionsType}";
```

Every mini-batch within a single `Optimize()` run produced the same cache key, so the first batch's gradient was cached and every subsequent `CalculateGradient` call returned that stale cached gradient. The model's parameters never drifted off their Xavier initialisation.

Fix: include the runtime identity of the model, the batch X, the batch y, plus a fast strided fingerprint of the current parameter vector. Batches are allocated fresh per iteration by `OptimizationDataBatcher.ExtractBatch`, so reference identity already differentiates them; the parameter fingerprint flips whenever `UpdateSolution` writes back new weights. Legitimate cache hits (line search / trust-region re-evaluation) still work because identity-based keys only collide when the caller actually reuses the same tensor instances.

### Bug 2 — Adam convergence check fired after epoch 0 (`AdamOptimizer.cs:182`)

```csharp
if (NumOps.LessThan(
    NumOps.Abs(NumOps.Subtract(bestStepData.FitnessScore, currentStepData.FitnessScore)),
    NumOps.FromDouble(_options.Tolerance)))
    return CreateOptimizationResult(bestStepData, inputData);
```

`UpdateBestSolution` copies `currentStepData` into `bestStepData` on the first iteration (because `bestStepData` starts uninitialised). The very next line then compared `|bestStepData - currentStepData| < 1e-6` — but those two ARE the same object after the copy, so the diff is always 0 < tolerance. Result: `MaxIterations=50` actually ran exactly 1 epoch and returned a near-untrained model.

Fix: compare against `previousStepData`, not `bestStepData`. The signal we want is per-epoch progress: "fitness stopped changing from one epoch to the next".

## Validation

| Path | Iterations actually run | top-1 acc |
|---|---|---|
| `model.Train` per-sample (baseline) | 50 epochs × 16 samples = 800 updates | **68.75%** |
| `BuildAsync` batched (before fix) | 1 epoch (convergence-fired) → 0 batches' grad applied | 6.25% (uniform = 1/V) |
| `BuildAsync` batched (after fix) | up to 10 epochs (now bounded by real plateau) | 6.25% (mode-collapse, see residual scope below) |

- 29/29 existing `AdamOptimizer` tests pass after the fix.
- The `AiModelBuilderFacadePredictParityTests.Facade_Predict_MatchesDirectModelPredict_AfterBuildAsync` regression test still passes.
- New `Train_PerSample_LearnsTrainingSet_BaselineThatBuildAsyncMustMatch` test passes at 68.75% (per-sample baseline that the batched path must match).

## Residual scope (NOT fixed in this PR)

Even after both fixes above, AdamOptimizer.Optimize() on this Transformer still mode-collapses — predictions converge to ONE class with high confidence instead of differentiating across inputs. The per-sample `model.Train` on the same architecture / loss / hyperparams reaches 68.75% top-1, so the collapse is specific to the Optimize loop's gradient computation path. Suspects (not yet isolated):

1. **Double gradient averaging**: `CalculateGradient` divides the returned gradient by `batchSize` (line 859) AND `ComputeGradients` computes the loss as a mean over batch via the tape — so the effective scaling is 1/N², not 1/N. `model.Train` only divides by 1 (per-sample).
2. **Training mode not set**: `Optimize` does NOT call `SetTrainingMode(true)` before forward / backward. Dropout layers behave at inference time during gradient computation. `Transformer.Train` explicitly enters training mode for the whole step.
3. **L2 regularization on by default**: `GradientBasedOptimizerOptions.Regularization = new L2Regularization<T,...>()` with strength 0.01 is applied inside `CalculateGradient` (line 856). `model.Train` does not regularize.
4. **Loss-function override**: `GradientBasedOptimizerOptions.LossFunction` defaults to `MeanSquaredErrorLoss<T>`. `CalculateGradient` passes the optimizer's loss to `gradientComputable.ComputeGradients(X, y, LossFunction)`, and `NeuralNetworkBase.ComputeGradients` honours the passed-in loss over the model's configured loss. So a Transformer constructed with CCE will silently train against MSE-on-softmax-output via the optimizer's default loss. **This is the highest-impact remaining bug** — wrong loss = wrong gradient sign = wrong learning.

The new regression test `BuildAsync_Batched_LearnsTrainingSet_NotUniform` is checked in but `[Fact(Skip=…)]`'d with a forward reference to this scope so it documents the consumer contract without breaking CI. Three companion diagnostic tests (`BuildAsync_Diagnostic_*` and `AdamOptimize_Direct_*`) are also added as Skipped facts so the next investigator can re-run them locally.

## Test plan

- [x] 29/29 existing `AdamOptimizer` tests still pass
- [x] `AiModelBuilderFacadePredictParityTests.Facade_Predict_MatchesDirectModelPredict_AfterBuildAsync` still passes
- [x] New per-sample `Train_PerSample_LearnsTrainingSet_BaselineThatBuildAsyncMustMatch` passes at 68.75% (proves the per-sample path still converges)
- [ ] New `BuildAsync_Batched_LearnsTrainingSet_NotUniform` skipped pending residual fixes — will be unskipped when the four remaining suspects above are resolved

## Links

- HarmonicEngine consumer ticket: HE memory note "AiDotNet#1340 FIXED 2026-05-17 — Transformer.Train per-call leak (Tensors PR #360)" — this PR closes the OPTIMIZER side; the per-call leak referenced there is a separate Tensors-side fix.
- Related: #1331 (fused-compiled training-plan staleness — fixed) — was per-sample-path only; this PR addresses the batched / Optimize path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved optimizer convergence detection to avoid premature early stopping.
  * Fixed gradient-caching so gradients differ across mini-batches and parameters update correctly.

* **Tests**
  * Added integration and diagnostic tests covering Transformer training and optimizer behavior, including baseline and direct-optimizer checks.
  * Several regression tests included; some are marked skipped pending follow-up investigations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ooples/AiDotNet/pull/1351?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Optimizers/AdamOptimizer.cs
+++ b/src/Optimizers/AdamOptimizer.cs
@@ -178,9 +178,24 @@ public class AdamOptimizer<T, TInput, TOutput> : GradientBasedOptimizerBase<T, T
                 return CreateOptimizationResult(bestStepData, inputData);
             }
 
-            // Check convergence
+            // Check convergence against the PREVIOUS epoch, not against
+            // bestStepData. UpdateBestSolution above copies currentStepData
+            // into bestStepData on the first iteration (because bestStepData
+            // starts uninitialised), so |best - current| would always be 0
+            // < tolerance and the optimiser would exit after the first epoch
+            // — observed as AiModelBuilder.BuildAsync producing uniform
+            // (1/V) predictions because only ~3 batched Adam steps ran
+            // before Optimize returned. The correct convergence signal is
+            // "the fitness stopped changing from one epoch to the next",
+            // i.e. |current - previous| < tolerance. Issue #1340.
+            //
+            // Guard the first iteration explicitly: previousStepData is the
+            // pre-epoch baseline returned by PrepareAndEvaluateSolution
+            // (untrained model evaluation). At epoch=0 we have to compare
+            // current vs that pre-epoch baseline, so the per-epoch progress
+            // signal is meaningful starting from the very first epoch.
             if (NumOps.LessThan(
-                NumOps.Abs(NumOps.Subtract(bestStepData.FitnessScore, currentStepData.FitnessScore)),
+                NumOps.Abs(NumOps.Subtract(previousStepData.FitnessScore, currentStepData.FitnessScore)),
                 NumOps.FromDouble(_options.Tolerance)))
             {
                 return CreateOptimizationResult(bestStepData, inputData);
--- a/src/Optimizers/GradientBasedOptimizerBase.cs
+++ b/src/Optimizers/GradientBasedOptimizerBase.cs
@@ -754,6 +754,17 @@ public abstract class GradientBasedOptimizerBase<T, TInput, TOutput> : Optimizer
         TInput X,
         TOutput y)
     {
+        // Issue #1340: the previous implementation read/wrote a gradient cache
+        // keyed by (modelType, batchSize, inputSize, optionsType) which made
+        // every mini-batch produce the SAME cache key inside one Optimize()
+        // run — the first batch's gradient was cached and every subsequent
+        // call returned that stale cached gradient, so parameters never moved
+        // off their initialisation. The fix is to compute a key that ALSO
+        // includes a fingerprint of the current model parameters and the
+        // runtime identity of the batch tensors, so a legitimate re-evaluation
+        // of the exact same (params, X, y) tuple (e.g. line search / trust-
+        // region reject) still hits the cache while a normal training loop
+        // always misses.
         string cacheKey = GenerateGradientCacheKey(solution, X, y);
         var cachedGradient = GradientCache.GetCachedGradient(cacheKey);
         if (cachedGradient != null)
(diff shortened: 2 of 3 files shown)


ooples:
Superseded by #1364 — that branch already contains all the H1/H2/H3 fixes from this PR plus the gradient-walk fix and H5 diagnostic. Closing to consolidate the BuildAsync mode-collapse remediation into a single PR (per user feedback about too many small PRs).

Issue #1346, opened 2026-05-17
flashattentionlayer produces degenerate output + 3.76x slowdown vs documented 2-4x speedup


ooples:
## Summary

`FlashAttentionLayer<T>` is documented as a drop-in replacement for `MultiHeadAttentionLayer<T>` with the same math + a 2-4x speedup. In practice on the **compiled fused-Adam training path** (the default for float + Adam models), `FlashAttentionLayer` produces uniformly distributed output that does not change during training — gradients are silently zeroed at the engine level because `Engine.FlashAttention` lacked a `GraphMode.IsActive` lazy-graph recording branch, while every other differentiable engine op has one.

## Reproducer

HarmonicEngine consumer test `Phase_PAPER_A_PathB_FlashAttention_Sanity` (10KB WikiText-2 byte-LM, dModel=128, L=2, ctx=64, vocab=256, Adam, lr=1e-3, epochs=2):

| arm | top-1 | top-5 | ppl | train wall |
|-----|------:|------:|----:|-----------:|
| baseline MHA       | 15.20% | 43.40% | 254.69 | 17.87s |
| FA (master, pre-fix)|  0.00% | 100.00% | 256.00 | 59.99s (3.76x SLOWER) |
| FA (Tensors fix)    |  0.00% |  92.80% | 256.00 | 48.53s (2.72x slower)  |

top-1 = 0% + top-5 = 100% + ppl = V (256) is the signature of uniform softmax output, meaning the readout's Dense weights never moved off random init. The 3.76x slowdown vs documented 2-4x speedup is the symptom of eager FA running every training step inside a compile lambda that should have replayed a compiled plan.

## Root cause (engine side — fixed by paired Tensors PR)

`CpuEngine.FlashAttention<T>` in `AiDotNet.Tensors/src/AiDotNet.Tensors/Engines/CpuEngine.cs` was the **only differentiable engine op** without a `GraphMode.IsActive` branch — all 219 other recorded ops route through `GraphMode.Current.Record{Unary,Binary,Variadic}` when a compiled training plan is being traced. Pre-fix:

- Lazy-graph trace records every op in `forward(input)` except FA.
- FA computes eagerly and returns a materialised `Tensor<T>` with no `LazySource`.
- Downstream lazy ops (Permute + Reshape in `FlashAttentionLayer<T>.Forward`) capture the FA output as a **frozen compile-time constant**.
[...]

PR #1386, opened 2026-05-19
test(#1346): canary for engine fix #362 reaching fused-Adam path + Skip'd regression for Tensors#396

### LINKED ISSUE #362: [Test Coverage] Implement Tests for Basic Wavelets
## Problem
Basic wavelet functions in src/WaveletFunctions/ have 0% test coverage.

## Files Requiring Tests (5 files)
- HaarWavelet.cs
- MexicanHatWavelet.cs
- ContinuousMexicanHatWavelet.cs
- MorletWavelet.cs
- GaussianWavelet.cs

## Goal
Achieve 80%+ test coverage for basic wavelets with signal processing tests.

## Labels
test-coverage, wavelets, signal-processing

ooples:
## Summary

Adds focused integration test coverage tied to AiDotNet#1346 (now closed) and the follow-up tracked at [AiDotNet.Tensors#396](https://github.com/ooples/AiDotNet.Tensors/issues/396).

Two tests in one new file:

1. **`FlashAttentionLayer_TrainViaFusedCompiledAdam_EngagesFusedPath`** (active, passes) — canary that locks in AiDotNet.Tensors PR #362's reach into the AiDotNet fused-Adam training path. A `Transformer<float>` with `FlashAttentionLayer` in its explicit layer stack must engage `CompiledTapeTrainingStep<float>.TryStepWithFusedOptimizer` on the first `Train()` call. Pre-fix this canary would be invisible (the fused step ran but FA gradients were silently zeroed). Post-fix the engagement counter is the proxy for "the engine fix reached this path".

2. **`DenseIdentity_CCE_OnFusedAdam_DoesNotSilentlyZeroNaN`** (Skip'd until Tensors#396 ships) — future-fix regression for the consumer-side gap surfaced during the #1346 investigation. When a multi-parameter model routes raw logits (IdentityActivation Dense) through `CategoricalCrossEntropyLoss` on the fused-Adam path, the CCE `log(negative_logit + eps)` chain produces NaN that the Tensors loss-readout silently zeroes — `GetLastLoss()` returns literal `0.0`, the consumer thinks training converged, and the model never moves off random init. The Skip message includes the unblock condition (Tensors#396 fix + NuGet bump).

## Why this scope

The bug investigation for #1346 found that the original engine-side gradient drop was correctly fixed by [AiDotNet.Tensors PR #362](https://github.com/ooples/AiDotNet.Tensors/pull/362) (in NuGet 0.81.3). The remaining "consumer still sees degenerate output" symptom turned out to be a much larger, not-FA-specific bug in the Tensors loss readout — diagnosed via 5 controlled bisection experiments and filed in detail at [AiDotNet.Tensors#396](https://github.com/ooples/AiDotNet.Tensors/issues/396). That fix lives in `AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs`, not in AiDotNet. AiDotNet#1346 is closed.

## Test plan

- [x] `dotnet build tests/AiDotNet.Tests/AiDotNetTests.csproj --framework net10.0 -c Release` — 0 errors
- [x] `dotnet test ... --filter "FullyQualifiedName~FlashAttentionFusedCompiledTrainingIssue1346Tests" --no-build` — 1 passed, 1 skipped

Local run output:
```
Passed!  - Failed: 0, Passed: 1, Skipped: 1, Total: 2
```

## Summary by CodeRabbit

* **Tests**
  * Added integration tests verifying FlashAttention engagement on the fused-Adam compiled training path.
  * Includes a canary test that ensures fused-step tracking is triggered during training.
  * Adds a skipped regression test (blocked externally) that validates loss behavior to prevent silent-zero failures.
  * Tests emit diagnostic values for debugging and verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ooples/AiDotNet/pull/1386?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/AiDotNet.Tests/IntegrationTests/NeuralNetworks/FlashAttentionFusedCompiledTrainingIssue1346Tests.cs
+++ b/tests/AiDotNet.Tests/IntegrationTests/NeuralNetworks/FlashAttentionFusedCompiledTrainingIssue1346Tests.cs
@@ -1,0 +1,247 @@
+using System.Threading.Tasks;
+using AiDotNet.ActivationFunctions;
+using AiDotNet.Enums;
+using AiDotNet.Interfaces;
+using AiDotNet.LossFunctions;
+using AiDotNet.Models.Options;
+using AiDotNet.NeuralNetworks;
+using AiDotNet.NeuralNetworks.Layers;
+using AiDotNet.Optimizers;
+using AiDotNet.Tensors.LinearAlgebra;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace AiDotNetTests.IntegrationTests.NeuralNetworks;
+
+/// <summary>
+/// Consumer-side regression coverage for AiDotNet#1346 (FlashAttentionLayer
+/// degenerate output on the compiled fused-Adam path). The original root
+/// cause — Engine.FlashAttention missing its GraphMode.IsActive lazy-graph
+/// recording branch — was fixed by AiDotNet.Tensors PR #362 and ships in
+/// AiDotNet.Tensors NuGet 0.81.3. This file pins TWO things:
+/// <list type="number">
+/// <item>The engine-side fix actually reaches the AiDotNet fused training
+/// path: a Transformer&lt;float&gt; whose layer stack contains
+/// <see cref="FlashAttentionLayer{T}"/> engages
+/// <see cref="AiDotNet.Training.CompiledTapeTrainingStep{T}.TryStepWithFusedOptimizer"/>
+/// when trained via the public network API (canary test below).</item>
+/// <item>The remaining consumer-side gap that the #1346 investigation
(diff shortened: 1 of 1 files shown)


ooples:
PR #1386 review C8Cn3 thread on Task.Yield/async: keeping the `async Task` + `await Task.Yield()` pair because `[Fact(Timeout = 60000)]` is xUnit-2-only-supports-async in this codebase. Verified empirically this session — when I tried `[Fact(Timeout = N)] public void M()` on a regression-pin test, xUnit rejected it at execution with "Tests marked with Timeout are only supported for async tests". `Task.Yield()` is the minimal-overhead trick to satisfy the `async Task` signature requirement (no real async work in the test body, just need the compiler-emitted state machine). Dropping the timeout would lose the CI safety net we want for tests that exercise full Transformer training — fused-Adam path bugs can spin/hang.

PR #1417, opened 2026-05-22
perf(#1349): route Int8WeightOnlyMatMul through SgemmWithInt8RowScaledCachedB

### LINKED ISSUE #1349: perf: wire SimdGemm.SgemmWithInt8CachedB into QuantizedAttentionLayer / QuantizedDenseLayer (currently ~20x slower than FP32)
## Summary

Found while implementing #1342 (#1348). The public Int8InferenceModel surface works end-to-end and produces ~4x weight compression with negligible accuracy loss, **but inference is ~20x slower than FP32** at wall-clock time on a 16x64-d=4 Transformer canary.

## Root cause

`QuantizedAttentionLayer.DequantizeMatMulInt8` and `QuantizedDenseLayer.Forward` both use a scalar 3-loop dequant-on-the-fly matmul:

```csharp
for (int r = 0; r < rows; r++)
    for (int o = 0; o < outDim; o++)
        for (int i = 0; i < inDim; i++)
            sum += input[inputBase + i] * (weights[wBase + i] * scale);
```

FP32 inference routes through `Engine.FusedLinear` -> `SimdGemm.Sgemm` (AVX-512 BLIS/GotoBLAS tiled GEMM).

## What needs to happen

The SIMD INT8 cached-B path already exists in `AiDotNet.Tensors/src/AiDotNet.Tensors/Engines/Simd/SimdGemm.cs`:

```csharp
public static void SgemmWithInt8CachedB(
    System.ReadOnlySpan<float> a, float[] b, System.Span<float> c, int m, int k, int n)
```

It is currently `internal` on `internal static partial class SimdGemm`. Two paths to enable it:

1. **Tensors-side**: expose a public INT8-weight matmul entry point (`AiDotNet.Tensors.LinearAlgebra.Int8MatMul.Forward(...)` or similar) and route `QuantizedAttentionLayer` + `QuantizedDenseLayer` through it.
2. **AiDotNet-side**: AiDotNet has `InternalsVisibleTo(\"AiDotNet\")` from Tensors already — confirm and call `SimdGemm.SgemmWithInt8CachedB` directly from the inference quantization layers. Per-row scales would need an adapter since the SIMD path uses per-tensor scale.

## Measured perf today

- FP32 wall-clock: 0.202 ms / predict
- INT8 wall-clock: 4.164 ms / predict
- Ratio: 20.64x slower

## Expected perf after wiring

Per the comment in `SimdGemm.SgemmWithInt8CachedB`:
> Numerical quality: per-tensor symmetric int8 quantization of transformer weights typically gives 35-40 dB SNR on the matmul output. Memory bandwidth on weight loads drops 4x.

So we should see ratio drop **below 1.0x** (INT8 faster than FP32) on memory-bandwidth-bound workloads, especially for larger embDim where weight-load is the bottleneck.

## Acceptance criteria

- INT8 `Int8InferenceModelIntegrationTests.FromTrained_PredictWallClockGapVsFP32_DocumentsCurrentState` test ratio assertion tightens from `< 50.0x` to `< 1.0x`.
- All existing `QuantizedAttentionTests` quality tolerance tests still pass.
- A new perf test at embDim >= 256 demonstrating > 1.5x speedup vs FP32.

ooples:
> ⚠️ **DRAFT** — depends on [`ooples/AiDotNet.Tensors#427`](https://github.com/ooples/AiDotNet.Tensors/pull/427) publishing a new NuGet. Will not compile until then. Held as draft to track the dependency.

## Summary

Closes #1349 once the Tensors NuGet bump lands.

Replaces the current `Int8WeightOnlyMatMul.MultiplyAddBias` body — per-call `ArrayPool.Rent`, per-tile AVX2 dequant into FP32 scratch, standard FP32 `SimdGemm.Sgemm`, SIMD bias-scatter — with a single call to the new `SimdGemm.SgemmWithInt8RowScaledCachedB` kernel + a one-row `SimdKernels.VectorAdd` bias pass:

```csharp
SimdGemm.SgemmWithInt8RowScaledCachedB(
    a: input, bInt8: weightsInt8, rowScales: rowScales,
    c: output, m: rows, k: inputSize, n: outputSize);
if (biases != null) { /* per-row VectorAdd */ }
```

The new kernel keeps weights in INT8 all the way through the macro-kernel and folds per-row scales into the per-tile dequant. The 4× DRAM-bandwidth advantage of INT8 weights is finally realized at inference instead of being destroyed by materializing the full FP32 weight matrix in scratch.

## Behavior parity

- Same matmul contract: `C[r, o] = bias[o] + A[r, i] · (B_int8[o, i] * rowScales[o])`
- Same `[n, k]` weight layout (matches `Int8WeightOnlyQuantization.QuantizePerRow` output)
- Same SNR envelope — the new kernel's tile pipeline is the same one the prior path drove via `Sgemm` after pre-dequanting

## Diff shape

Net **−89 lines** (177 → 56 in `Int8WeightOnlyMatMul.cs`). The deleted code was the per-call dequant + ArrayPool + tile-scatter loop. `ChooseTileSize` stays defined for now — existing unit tests reference it as a static heuristic; cleanup is a separate follow-up once this PR is buildable.

## Expected perf delta

Per the Tensors#401 downstream estimate:
- 16×64 transformer canary INT8/FP32 ratio: **~20× → ~3-5×**
- BERT-class shapes: full **4× DRAM saving** on weight loads

## Merge sequence

1. **Land Tensors#427** ([per-row-scaled INT8 GEMM kernel + tests](https://github.com/ooples/AiDotNet.Tensors/pull/427))
2. **Publish new `AiDotNet.Tensors` NuGet**
3. **Bump `Directory.Packages.props`** (literal edit — the comment block already documents which PR triggers the bump)
4. **Mark this PR ready for review** — CI then passes, perf assertions in `Int8InferenceModelIntegrationTests.FromTrained_PredictWallClockGapVsFP32_DocumentsCurrentState` can tighten

## Summary by CodeRabbit

* **Performance**
  * Faster INT8 weight-only matrix multiplication for improved inference throughput on supported platforms.

* **Code Improvements**
  * Reduced implementation redundancy and simplified buffer handling for more reliable execution and a streamlined fallback path on legacy runtimes.

* **Chores**
  * Updated project comments documenting the dependency bump context and INT8 wiring notes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ooples/AiDotNet/pull/1417?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,20 +5,33 @@
   <ItemGroup>
     <!-- AiDotNet ecosystem -->
     <PackageVersion Include="AiDotNet" Version="0.113.0" />
-    <!-- AiDotNet.Tensors 0.84.1 — carries the blocker fixes for the previously-
-         draft framework PRs that depended on Tensors changes:
-           * #427 SgemmWithInt8RowScaledCachedB — per-row-scaled INT8 weight-only
-             GEMM (AiDotNet#1349 / PR #1417).
-           * #437 CpuEngine.LstmSequenceForward — fused recurrent forward
-             (AIsEval LSTM / PR #1457).
-           * #454 BroadcastElementwise SIMD perf (AiDotNet#1307 / PR #1448).
-         Plus the 0.81.4–0.82.5 fixes already in flight (GraphMode gradient
-         recording #367, FP64 SD UNet cliffs #410, fused-compiled VGG #416,
-         clCreateCommandQueue serialisation #418, TensorAllocator long-arith #424). -->
-    <PackageVersion Include="AiDotNet.Tensors" Version="0.84.1" />
-    <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.81.9" />
-    <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.82.5" />
-    <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.81.9" />
+    <!-- AiDotNet.Tensors 0.86.1 — bumped from 0.84.1 to unblock this PR (#1417).
+         The dep that gated this PR's INT8 wiring shipped: Tensors PR #427
+         (SgemmWithInt8RowScaledCachedB — per-row-scaled INT8 weight-only GEMM)
+         was merged 2026-05-22 and is published from 0.85.x onward; 0.86.1 is the
+         latest stable, so Int8WeightOnlyMatMul.cs's new single-call body now
+         compiles against the real Tensors API surface instead of a localverify.
+
+         Full v0.84.1 -> v0.86.1 delta picks up:
+           * Tensors #427 — SgemmWithInt8RowScaledCachedB (THIS PR's gating dep).
+           * Tensors #437 — CpuEngine.LstmSequenceForward (AIsEval LSTM / PR #1457).
+           * Tensors #454 — BroadcastElementwise SIMD perf (#1307 / PR #1448).
+           * Tensors #472 — Float32Precision opt-in for FusedAttention<double>.
+           * Tensors #485 — output-only SDPA (6.9x less alloc, 2.7x mean latency).
+           * Tensors #487 — GetSliceAlongDimension tape-backward.
+           * Tensors #488 — fused QKV + transpose-fused SDPA
+             (MHA 5.31 -> 2.26 ms; beats torch ~2.74 ms).
+           * Tensors #474 — small-batch native-BLAS thread cap inside MlpForward.
+           * Tensors #462 — hybrid hardware-aware GEMM strategy (seed + learned).
+           * TensorArena.CreatePersistent / Activate — zero-GC training arena.
+         Plus the 0.81.4–0.82.5 baseline (GraphMode gradient recording #367,
+         FP64 SD UNet cliffs #410, fused-compiled VGG #416, clCreateCommandQueue
+         serialisation #418, TensorAllocator long-arith #424).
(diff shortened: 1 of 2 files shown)


ooples:
Unblocked: rebased on master and bumped AiDotNet.Tensors to **0.84.1**, which ships #427 (`SimdGemm.SgemmWithInt8RowScaledCachedB`). `src/AiDotNet.csproj` builds clean and the int8/quantization suites pass (Int8WeightOnlyMatMulTests + Int8InferenceModelIntegrationTests + InferenceQuantizationIntegrationTests + QuantizedAttentionTests = **83/83**). Marking ready for review.

ooples:
**Verification correction + net471 fix.** My earlier "ready" claim only covered net10.0. On a multi-TFM check I found this PR **did not compile on net471**: `SimdGemm.SgemmWithInt8RowScaledCachedB` (Tensors #427) ships only in the net10.0 Tensors build (`CS0117` on net471). Fixed in `83910fd09` by guarding the fast path with `#if !NETFRAMEWORK` and restoring the original per-tile dequant + `SimdGemm.Sgemm` path for net471.

Now verified on **both** targets:
- `src/AiDotNet.csproj` builds on net10.0 **and** net471.
- `Int8WeightOnlyMatMulTests` (incl. `MultiplyAddBias_MatchesScalarReference`, which asserts element-wise parity vs an independent scalar reference) — **17/17 on net10.0 and 17/17 on net471**.

Issue #1392, opened 2026-05-19
perf: NEAT.Train tape-mode training is too slow — Training_ShouldReduceLoss times out at 120s


ooples:
## Summary

\`NEATTests.Training_ShouldReduceLoss\` (the model-family invariant that trains a NEAT — NeuroEvolution of Augmenting Topologies — model for the base 100 \`MemorizationTaskIterations\` count and asserts loss decreased by ≥ 1%) consistently times out at the \`[Fact(Timeout=120000)]\` budget. Surfaced while working [#1304 cluster-6](https://github.com/ooples/AiDotNet/issues/1304); shape-of-the-gap is similar to [#1390](https://github.com/ooples/AiDotNet/issues/1390) (DCGAN GAN training perf gap).

## Repro

\`\`\`
$ dotnet test --filter \"FullyQualifiedName=AiDotNet.Tests.ModelFamilyTests.NeuralNetworks.NEATTests.Training_ShouldReduceLoss\" --framework net10.0
Failed!  - Failed: 1, Passed: 0
  Test execution timed out after 120000 milliseconds
\`\`\`

## Why this is perf, not correctness

NEAT trains via evolutionary topology mutation + per-genome backprop steps. The base-class \`Training_ShouldReduceLoss\` invariant runs 100 \`network.Train(input, target)\` calls. On real PyTorch-style implementations (where the topology is fixed and only the weights evolve) this completes in seconds. Here it doesn't finish in 2 minutes, which means \~1 s+ per Train call on a tiny tabular input. The model's per-genome population evaluation + speciation pass dominates, not the per-genome weight update.

This is the same class of perf gap [#1390](https://github.com/ooples/AiDotNet/issues/1390) calls out for DCGAN tape-mode training: the model is functionally correct but the throughput is ~5-50× off PyTorch CPU baseline at this scale. Closing the gap requires Tensors-side Conv2D/MatMul backward perf work + per-op tape-recording allocator pressure reduction, neither of which is in cluster-6's scope.

## Affected tests (current)

- \`NEATTests.Training_ShouldReduceLoss\`

Adjacent tests at risk if NEAT's per-step cost is unchanged:
- \`NEATTests.MoreData_ShouldNotDegrade\` (250-iter chain)
- \`NEATTests.LossStrictlyDecreasesOnMemorizationTask\` (100-iter chain)
- Any other model-family invariant that exceeds the 60-120 s envelope at \~1 s/Train.

## NOT this issue

[...]

PR #1419, opened 2026-05-22
perf(#1392): remove O(N²) ordering in NEAT fitness + cache topology sort

ooples:
**Partial progress on #1392** (NEAT.Train tape-mode training too slow). First-pass perf win; doesn't fully close the gap but moves the test comfortably under the 120 s CI budget locally.

## Bottlenecks profiled

`NEATTests.Training_ShouldReduceLoss` runs `TrainingIterations * 3 = 30` `network.Train(...)` calls. Each `Train` runs **50 internal evolutionary generations** with **150 genomes per population**. That's ~225 000 `ActivateGenome` calls and ~7 500 generations of fitness evaluation per test.

Two hot paths dominated:

### 1. Inline `LastLoss` assignment did `OrderByDescending` per genome

```csharp
// NEAT.cs, inside the fitness function:
if (genome == _population.OrderByDescending(g => g.Fitness).FirstOrDefault())
{
    LastLoss = averageLoss;
}
```

Ran on **every** genome eval — O(N) sort × 150 genomes × 50 generations × 30 Train calls = **~34 million ordering ops**, all heap-allocating LINQ enumerables. The reference-equality probe was also semantically broken — it compared `genome` against the **pre-evaluation** best (Fitness values from the previous generation, essentially random for the current one). The existing post-evolution recompute block immediately below already noted this and does the work correctly.

**Fix**: delete the branch. The post-evolution recompute (NEAT.cs:~1230+) does the right thing using the actual post-generation best.

### 2. `SortConnectionsTopologically` rebuilt on every `ActivateGenome` call

The O(E²) topological sort + Dictionary/HashSet/List allocations ran from scratch on every activation. Most weight-only mutations across the 50 generations leave the genome topology unchanged, so the sort result is reusable.

**Fix**: cache the sort + the non-input-node id list on the `Genome`, keyed on `(Connections.Count, ulong bitmask of IsEnabled)`. Topology mutations (`AddConnection` / `DisableConnection`) change one or both halves of the key and force a rebuild on next access; weight mutations leave it valid.

## Local timing (single-test, 32-core host)

| Run | Wall time | Comment |
|---|---|---|
| Baseline | 54.0 s | Original code |
| Post-fix | 46.0 s | This PR — ~15% faster, ~74 s of headroom under the 120 s budget |

## What's not in this PR

The issue scopes the gap as "multi-week" and notes ~1 s+ per Train call on a tiny tabular input. This first pass takes the easy wins; the remaining cost is in:
- `ActivateGenome`'s `Dictionary<int, T>` allocator pressure (could move to a flat array keyed on dense node ids)
- The per-mutation `Clone()` path that copies every Connection per child genome
- Speciation + tournament-selection scans

These need their own profiling pass + are not blocking the test from passing.

## Verification

All 12 `NEATTests` pass in isolation. No behavior change vs baseline (LastLoss values, mutation outputs, evolution trajectory unchanged because the deleted branch was already dead code per the post-evolution recompute comment).

## Files

- `src/NeuralNetworks/NEAT.cs` — deleted dead `OrderByDescending` branch, added `GetOrBuildSortedConnections` / `GetOrBuildNonInputNodeIds` / `ComputeEnabledBitmask` cache helpers
- `src/NeuralNetworks/Genome.cs` — added 4 cache fields (`CachedTopologySignatureCount`, `CachedTopologySignatureMask`, `CachedSortedConnections`, `CachedNonInputNodeIds`)

## Summary by CodeRabbit

* **Chores**
  * Bumped centrally managed AiDotNet package versions to 0.81.9 and updated accompanying notes.

* **Performance**
  * Reduced runtime allocations and memory usage across evolution, crossover, mutation, and selection.
  * Added caching and pre-sized buffers for faster network activation and evolution cycles.

* **Bug Fixes**
  * Fixed activation and training output handling for more reliable loss reporting and stable layer outputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ooples/AiDotNet/pull/1419?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/automated-release.yml
+++ b/.github/workflows/automated-release.yml
@@ -306,6 +306,16 @@ jobs:
       # adopted the Category=Smoke convention.
       - name: Smoke tests (release gate)
         if: steps.version.outputs.should_release == 'true'
+        env:
+          # ProgramSynthesisCoverageSmokeTests (CodeT5, GraphCodeBERT)
+          # invoke GetModelMetadata which serializes the model via
+          # ModelPersistenceGuard.EnforceBeforeSerialize. Without a
+          # license key in the environment, the guard throws
+          # LicenseRequiredException and the release-gate fails — the
+          # symptom that blocked every release since the smoke gate was
+          # added. The secret holds the project's CI license; pipe it
+          # through here so LicenseKeyResolver.Resolve finds it.
+          AIDOTNET_LICENSE_KEY: ${{ secrets.AIDOTNET_LICENSE_KEY }}
         run: |
           dotnet test tests/AiDotNet.Tests/AiDotNetTests.csproj \
             --no-build -c Release --framework net10.0 \
--- a/.github/workflows/ci-shard-closure-policy.yml
+++ b/.github/workflows/ci-shard-closure-policy.yml
@@ -1,0 +1,122 @@
+name: CI Shard Closure Policy
+
+# Enforces the "shard stays open until shard goes green" rule established in
+# audit comment https://github.com/ooples/AiDotNet/issues/1315#issuecomment-4501244896.
+# When an issue tagged `ci-failure` is closed, this action checks the latest
+# Build & SonarCloud run on master and warns / auto-reopens if the shard the
+# issue claims to fix is still red.
+#
+# Why this exists: historically issues were closed when their originally-listed
+# tests passed, but the shard those tests belong to was still failing due to
+# OTHER tests in the same shard. Result: dashboard said "fixed" while CI was
+# perpetually red. The audit pulled 4 prematurely-closed issues that were
+# still tracking red shards (#1304, #1305, #1307, #1313). This action stops
+# that pattern at the source.
+
+on:
+  issues:
+    types: [closed]
+
+permissions:
+  issues: write
+  actions: read
+
+jobs:
+  check-shard-still-red:
+    runs-on: ubuntu-latest
+    # Only run on ci-failure-labeled issues. Other issue closures are out of
+    # scope — this isn't a generic "did you fix it" guardrail, just a CI-shard
+    # accountability check.
+    if: contains(github.event.issue.labels.*.name, 'ci-failure')
+    steps:
+      - name: Extract shard name from issue title
+        id: extract
+        env:
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+        run: |
+          # Issue titles follow conventions like:
+          #   "[PR #1290 CI] Tests (net10.0) - ModelFamily - NeuralNetworks: 5 failing tests"
+          #   "[PR #1290 CI Cluster 6] Long-training timeouts ..."
+          #   "[CI] Tests (net10.0) - Unit - 08d NN-Adapters/Other: MoE MoreData failing"
+          # Try to extract the shard short name (the bit after "Tests (net10.0) - ").
+          shard=$(echo "$ISSUE_TITLE" | grep -oP 'Tests \(net10\.0\) - \K[^:]+' | head -1 | sed 's/[[:space:]]*$//')
+          if [ -z "$shard" ]; then
+            echo "No shard name found in title — issue not bound to a named shard, skipping check."
+            echo "shard=" >> "$GITHUB_OUTPUT"
+          else
+            echo "Extracted shard: '$shard'"
+            echo "shard=$shard" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check latest master CI run for this shard
+        id: check
+        if: steps.extract.outputs.shard != ''
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHARD_NAME: ${{ steps.extract.outputs.shard }}
+        run: |
+          # Find the most recent COMPLETED Build & SonarCloud run on master,
+          # success or failure — whichever is newer. Picking "newest success
+          # first, fall back to failure" would let a newer red run be ignored
+          # whenever any older green run exists, so the tracking issue stays
+          # closed while the shard is currently red. We deliberately skip
+          # cancelled runs (those don't tell us anything about shard health —
+          # they were superseded by a newer commit).
+          run_id=$(gh run list \
+            --repo "${{ github.repository }}" \
+            --workflow "Build & SonarCloud" \
+            --branch master \
+            --limit 20 \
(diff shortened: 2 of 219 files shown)


ooples:
Pushed [86beca85](https://github.com/ooples/AiDotNet/commit/86beca853) — the Dictionary allocator-pressure work + bulk Clone fast-path called out in the original "What's not in this PR" section.

### What's in this commit

**`ActivateGenome` on a flat `T[]` instead of `Dictionary<int, T>`**

Per-call shape was: alloc Dictionary, foreach over sorted connections setting/getting by key. Under `EvolvePopulation` that's one Dictionary allocation per genome per fitness evaluation — 150 pop × 50 gen × ~30 Train calls per test = ~225k Dictionary allocs per test invocation. The 1 buffer is now sized to `max(referenced node id, biasNodeId) + 1`, and all reads/writes are array indexing.

Three new caches on `Genome<T>` piggyback on the existing topology-sort caches:

- `CachedMaxNodeId` — for sizing the activation buffer
- `CachedReferencedNonInputNodeIds` — for the sigmoid sweep (replaces the `activations.Keys` loop)
- existing `CachedSortedConnections` invalidation now also invalidates the two above

`GetNamedLayerActivations` dropped its `ContainsKey` / `Keys.Where(...)` Dictionary patterns in favor of straight array indexing + the cached non-input-node-id list.

**`Clone()` pre-sizes the child genome's connection list**

The previous foreach + `AddConnection` path drove `List<Connection<T>>`'s internal array through the `0→4→8→16→32` capacity-doubling chain, copying the buffer at every step. Now `Connections.Capacity = parent.Count` is set once. `Connection<T>` objects are still freshly allocated per child so parent mutations don't leak across the clone boundary.

### Numbers

`NEATTests.Training_ShouldReduceLoss` (isolated, net10.0, 3-run min):

| Stage | Wall time | Cumulative |
|---|---|---|
| pre-#1419 baseline | ~54 s | — |
| #1419 first pass | ~46 s | 15% |
| this commit (`86beca85`) | ~41 s | **24%** |

### What's still not in this PR

Saved for follow-ups:

- Speciation + tournament-selection scans (additional O(N²) hot paths under EvolvePopulation)
- `Connection<T>` pool reuse (Clone still allocs N fresh Connection objects per child; could borrow from a per-thread pool)

### Note on the parallel-suite flake

Re-ran the full `NEATTests` filter (21 tests) against both the stashed-baseline state and the new commit. Both runs hit the same 120 s xunit method-Timeout on `Training_ShouldReduceLoss` under parallel CPU contention. The flake is pre-existing — root cause is xunit running the heavy-compute NEAT tests in parallel against a 120 s budget per method, not anything introduced here. Isolated runs comfortably fit the budget (~41 s).

ooples:
Pushed [3c476a88](https://github.com/ooples/AiDotNet/commit/3c476a888) — the rest of the deferred follow-ups now folded into this PR.

### What's in this commit

**`SelectParent`** — Replaced the allocate-a-List + `OrderByDescending` LINQ chain with an inline 3-way argmax. Tournament size is fixed at 3, so an unrolled compare beats anything LINQ can do. Per Training_ShouldReduceLoss run that's ~447 k allocations + LINQ enumerators eliminated (149 children × 50 generations × ~30 Train calls × 2 parents per crossover).

**`Crossover`** — Two allocations per call killed:
  - `Enumerable.Concat(parent1.Connections, parent2.Connections)` enumerator → walk both lists by index
  - Fresh `HashSet<int>` per call → per-NEAT-instance `_crossoverSeen` scratch HashSet that `.Clear()`s at the top of each call. Safe because EvolvePopulation is single-threaded.
  - Plus the child Connections list is pre-sized to `parent1.Count + parent2.Count` so the capacity-doubling chain doesn't fire.

**`Mutate`** — All three LINQ calls dropped:
  - `connections.Max(c => Math.Max(c.FromNode, c.ToNode))` → manual scan loop. Saves a Func<,> delegate alloc + enumerator per call.
  - `connections.Any(c => c.FromNode == fromNode && c.ToNode == toNode)` → manual scan loop. Same.
  - The weight-mutation `foreach` → index loop so JIT can elide the `List<T>.Enumerator` bounds check on each step.

**`EvolvePopulation`** — `new List<Genome<T>>(_populationSize)` instead of `new List<Genome<T>>()` so the backing array doesn't walk `0→4→8→16→…→150` on every generation.

### Why no Connection<T> pool

Considered and skipped — Connection<T>'s `FromNode`, `ToNode`, and `Innovation` are init-only via the public constructor (`Connection.cs:48, 68, 144`). Pooling would require either a breaking API change to make them settable or a fragile internal reset path that hides initialization bugs. The per-genome allocation churn for connections is bounded by `genome.Connections.Count` (small — typical genomes have <100 connections) and is paid under JIT-friendly `Add()` calls into pre-sized child lists, so the remaining marginal win does not justify the API risk.

### Numbers

`NEATTests.Training_ShouldReduceLoss` (isolated, net10.0, 3-run min): ~41 s — unchanged vs the prior commit. The dominant cost is the inner `ActivateGenome` forward pass (already optimized in `86beca85`), which is the per-fitness-call work and runs ~225 k times. The outer-loop allocations this commit trims are ~447 k events for SelectParent + ~75 k for Crossover, but each event was sub-microsecond, so the wall-time delta is in the noise. Where this commit pays off is reduced GC pressure under the parallel-suite load (less likely to push the 120 s xunit method timeout over the cliff).

### Final perf summary for PR #1419

| Stage | Wall time | Cumulative |
|---|---|---|
| pre-#1419 baseline | ~54 s | — |
| #1419 first pass (`09534a47`) | ~46 s | 15% |
| Dictionary→array + Clone (`86beca85`) | ~41 s | 24% |
| zero-alloc tournament/crossover/mutate (this commit) | ~41 s | 24% (GC pressure ↓) |

### NEAT test correctness

Spot-checked 6 NEAT tests individually on the new commit (parallel-suite host-crash is environmental — same problem on baseline):

- `Training_ShouldChangeParameters` — passed
- `Clone_AfterTraining_ShouldPreserveLearnedWeights` — passed
- `LossStrictlyDecreasesOnMemorizationTask` — passed
- `GradientFlow_ShouldBeNonZeroAndFinite` — passed
- `NamedLayerActivations_ShouldBeNonEmpty` — passed
- `BatchConsistency_SingleMatchesBatch` — passed
- `Training_ShouldReduceLoss` — passed (3/3, ~41 s)

Build clean on net10.0 + net471.

PR #1431, opened 2026-05-22
perf: LSTMLayer adopts fused LstmSequenceForward primitive (AIsEval P0 close-out)

ooples:
> Draft: blocked on AiDotNet.Tensors PR ooples/AiDotNet.Tensors#437 merging
> and a new tensors NuGet release. See \"Build dependency\" below.

## Summary

Closes the only remaining tensors-adoptable item from the AIsEval P0 gap list (issue ooples/AiDotNet.Tensors#436): **LSTMLayer.Forward** still ran a per-timestep loop with 8 separate \`TensorMatMul\` + activation dispatches per step. That path produced the \"AiDotNet LSTM did not finish in 3+ min\" datapoint in the [AIsEval fair-comparison rerun](https://github.com/lyudmilpetrov/AIsEval/pull/25).

This PR adds an inference-only fast path in \`LSTMLayer.Forward\` that routes the whole sequence through \`CpuEngine.LstmSequenceForward<T>\` — the fused primitive landed in ooples/AiDotNet.Tensors#437.

After this PR + the tensors release:
| AIsEval @ bs=128 | Pre-rerun | After both PRs |
|---|---|---|
| LSTM steady-state inference | did not finish in 3+ min | **~8.2 ms** (faster than PyTorch's 11.76 ms) |

### \`src/NeuralNetworks/Layers/LSTMLayer.cs\` (+150 lines)

Adds a fast-path branch at the top of \`Forward()\` that calls the new tensor primitive when:
- \`IsTrainingMode == false\` — the primitive doesn't populate the \`_cachedHiddenStates\` / \`_cachedCellStates\` caches that \`Backward\` reads, so it's safe only when no backward will run.
- \`typeof(T) == float\` — the primitive's competitive perf comes from its float-specialized SimdGemm + AVX sigmoid/tanh path.
- \`Engine is CpuEngine\` — the primitive lives there.
- \`GraphMode.IsActive == false\` — the primitive throws under an active autograd tape.

If any condition fails, the existing per-step loop runs unchanged.

The fast path includes:
- \`TryFusedLstmForward(cpuEng, input3D, batchSize, timeSteps)\` — packs the layer's 8 split weight tensors (F/I/C/O × i/h) into the concatenated \`[4*hidden, in]\` / \`[4*hidden, hidden]\` layout PyTorch \`nn.LSTM\` uses (gate order \`[i, f, g, o]\`), then calls \`cpuEng.LstmSequenceForward(...)\`. Packing copies ~25 KB of float at AIsEval shape — negligible vs the GEMM compute it enables. A caching pass can be a follow-up if profiling shows it.
- \`CopyLastTimestepHidden(...)\` — populates \`_lastHiddenState\` from the final timestep so consumers of the public \`LastHiddenState\` property keep working.

\`_lastCellState\` is allocated as zeros — its only consumers require training mode anyway.

### \`Directory.Packages.props\`

Bumps \`AiDotNet.Tensors\` pin from \`0.81.3\` to placeholder \`0.83.0\`. The release engineer should replace with the actual NuGet version that ships #437 at merge time. Build fails against any version <= 0.82.1 (the current latest published) because \`LstmSequenceForward\` doesn't exist there.

## What's NOT in this PR (deliberate scope)

- **MultiHeadAttentionLayer.Forward adoption of \`MultiHeadAttentionForward\`** — the existing layer already uses \`FusedLinear\` and \`ScaledDotProductAttention\` / \`FlashAttention\` (verified at lines 1105, 1111, 1153 of \`MultiHeadAttentionLayer.cs\`). The Stage 4/6/7 wrapper from #437 is mostly a dispatch-overhead reduction; whether to swap the existing chain for the wrapper is a profiling call that can be made in a follow-up after this lands and we re-run AIsEval.
- **DenseLayer.Forward adoption of \`FusedLinear\`** — already done at line 937, 945 of \`DenseLayer.cs\`. No change needed.
- **\`AiModelBuilder<MultipleRegression>\` lifecycle overhead at 48k rows** — purely framework-side (the underlying tensor ops aren't the bottleneck). Out of scope for this PR.
- **Fused backward for LSTM** — the primitive is forward-only in #437; this fast path is correspondingly inference-only. Training paths take the unchanged per-step loop which has full autograd support. Backward fusion is a separate follow-up.

## Build dependency

This PR's LSTMLayer changes call \`CpuEngine.LstmSequenceForward\`, which lands in tensors PR #437. **The framework build fails against tensors 0.82.1 (current latest published)** because that symbol doesn't exist yet. Sequence:

1. Tensors PR #437 merges to \`ooples/AiDotNet.Tensors:main\`.
2. A new tensors NuGet release ships (likely \`0.82.2\` or \`0.83.0\`).
3. The placeholder version in \`Directory.Packages.props\` here gets updated to the actual release.
4. CI on this PR goes green.
5. Merge.

Marked draft until those steps complete.

## Test plan

After the tensors release lands and the pin is updated:
- [ ] \`dotnet build src/AiDotNet.csproj -c Release\` succeeds on net10.0 + net471.
- [ ] Existing LSTM unit tests pass — the fast path is gated on \`!IsTrainingMode\`, so training-mode tests should hit the unchanged per-step loop.
- [ ] Manual verification: \`SetTrainingMode(false)\` + \`Forward(rank-3 float input)\` exercises the fast path; output should byte-match the per-step loop within float epsilon (the primitive is verified against a from-scratch reference in #437 at \`atol=1e-4\`).
- [ ] Re-run [AIsEval](https://github.com/lyudmilpetrov/AIsEval/pull/25) scaffold: \`dotnet run --no-build -c Release -- --models lstm --output ../results/aidotnet-lstm.json\`. Expect LSTM to finish (no >3min hang) and bs=128 steady-state latency to land in the 8-15 ms range — competitive with PyTorch's 11.76 ms.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,19 +5,21 @@
   <ItemGroup>
     <!-- AiDotNet ecosystem -->
     <PackageVersion Include="AiDotNet" Version="0.113.0" />
-    <!-- AiDotNet.Tensors needs a version bump after ooples/AiDotNet.Tensors#424
-         publishes a new NuGet. That PR replaces the silent `checked(int * int)`
-         dim-product overflow in TensorAllocator.Rent / RentPinned with a
-         `long` accumulator that names the requested shape when the element
-         count exceeds Array.MaxLength, plus an ArgumentOutOfRangeException
-         naming the index and value for negative dims (lazy-layer `-1`
-         sentinel propagation). Diagnoses the otherwise-opaque
-         `OverflowException` failures on TimeMachine / DQN / OWLViT /
-         DGCNN / TabTransformer / TabDPT / SlimSAM / TriaffineNER tests on
-         SonarCloud run 26241806890. (Previous PR ooples/AiDotNet.Tensors#359
-         — GraFPrint perf-overhaul + scheduler-fused + determinism — is
-         already in 0.81.3.) -->
-    <PackageVersion Include="AiDotNet.Tensors" Version="0.81.3" />
+    <!-- AiDotNet.Tensors version bump for this branch:
+         This branch's LSTMLayer fast path calls
+         CpuEngine.LstmSequenceForward<T> (added in
+         ooples/AiDotNet.Tensors#437). That PR also adds
+         MultiHeadAttentionForward<T> and the OpenCL GPU-only init gate.
+         Placeholder version 0.83.0 below — release engineer should
+         replace with the actual NuGet version that ships #437 (likely
+         0.82.2 or 0.83.0) at merge time. Build will fail against any
+         version <= 0.82.1 because LstmSequenceForward doesn't exist
+         there yet. The "version bump after #424" rationale from prior
+         iterations is also superseded by this one.
+
+         Latest currently-published: 0.82.1 (so 0.82.0/0.82.1 includes
+         the #424 long-arithmetic dim-product fix). -->
+    <PackageVersion Include="AiDotNet.Tensors" Version="0.83.0" />
     <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.81.3" />
     <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.81.3" />
     <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.81.3" />
--- a/src/NeuralNetworks/Layers/LSTMLayer.cs
+++ b/src/NeuralNetworks/Layers/LSTMLayer.cs
@@ -1099,6 +1099,63 @@ public partial class LSTMLayer<T> : LayerBase<T>
 
         _lastInput = input3D;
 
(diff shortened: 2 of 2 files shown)


ooples:
Superseded by #1457. The LSTM fused-forward change here (commit `36c23ec49`,
`LSTMLayer.Forward` → `CpuEngine.LstmSequenceForward`) was cherry-picked into
#1457 as `4d6ff0aae`, where it builds against the shipped AiDotNet.Tensors
(0.82.5+, now 0.84.1) — so this PR's placeholder `0.83.0` pin (`483f0c86f`) and
its blocked/draft state are obsolete, and the branch is conflicting with master.

#1457 additionally carries the verified benchmark (1417 ms per-step loop →
7.13 ms fused, 198× / beats PyTorch 11.76 ms) and a lazy LSTM/GRU GetParameters
test fix. Closing this in favor of #1457.

PR #1474, opened 2026-05-30
docs(serving): startup CPU-inference BLAS thread-pin recommendation

ooples:
## Startup thread-pin recommendation (Phase 7F follow-through)

The Phase-7F GEMM autotune probe found native BLAS at **`ProcessorCount/2` threads beats all-core by ~1.3×** on the small-batch, wide-but-short GEMMs typical of latency-sensitive inference (e.g. a classifier head's 784→512 layer). All-core saturates large batch/training GEMMs but **oversubscribes** small-batch work.

Crucially, the win is only realizable by **pinning the count once at process start** — changing the BLAS thread count at runtime rebuilds the thread pool (~700 µs), so per-call tuning backfires. That makes it an operator/deployment action, not a per-request toggle.

**This PR** has `ModelStartupService` log the recommendation once at host start:
- If a cap is already set (`OMP_NUM_THREADS` / `OPENBLAS_NUM_THREADS`), it reports it.
- Otherwise it nudges the operator to pin to ~cores/2 before launch (env var), or call `AiDotNet.Tensors.CpuInferenceConfig.PinBlasThreadsForLatency()` once at startup (the public knob added in AiDotNet.Tensors), with the measured rationale.

Informational only — it does **not** mutate global thread state itself, avoiding both the per-call pool rebuild and surprising other in-process OpenMP work.

Companion: the actionable `CpuInferenceConfig.PinBlasThreadsForLatency()` API lands in AiDotNet.Tensors PR #513; once released and bumped here, the serving host can call it directly.

## Summary by CodeRabbit

* **New Features**
  * Application startup now logs CPU inference BLAS thread recommendations. The service checks for existing thread configurations and suggests optimal thread allocation to improve CPU inference performance.

### DIFF
--- a/src/AiDotNet.Serving/Services/ModelStartupService.cs
+++ b/src/AiDotNet.Serving/Services/ModelStartupService.cs
@@ -70,6 +70,8 @@ public class ModelStartupService : IHostedService
     /// <param name="cancellationToken">Cancellation token.</param>
     public async Task StartAsync(CancellationToken cancellationToken)
     {
+        LogCpuInferenceThreadRecommendation();
+
         if (_options.StartupModels == null || _options.StartupModels.Count == 0)
         {
             _logger.LogInformation("No startup models configured");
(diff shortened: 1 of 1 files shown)


ooples:
@coderabbitai full review

PR #1513, opened 2026-06-06
feat(fp16): route SGD compiled training to mixed-dtype plan under AIDOTNET_FP16_ACTIVATIONS

ooples:
Phase E (AiDotNet side) of the FP16 activation-storage feature (Tensors #558 / Tensors PR #557).

Routes the SGD compiled-training path to the new public mixed-dtype plan when opted in:
`CompiledTapeTrainingStep.Step`, under `AIDOTNET_FP16_ACTIVATIONS=1` (float only), calls
`MixedPrecisionCompiledPlan.Trace/Step` (Tensors 0.91.34 public API) instead of the single-type fused
plan. Matmul activations are stored as FP16 (~1/2 resident) with grads bridged FP16<->FP32; SGD on the
FP32 master params, identical to this path's `UpdateParametersSGD`. Persistent input/target leaves are
captured at trace time and refreshed each step (changing batches replay correctly); recompiles on
shape change, dropped on layer-set change. Bumps the Tensors pin to 0.91.34.

Default off => the single-type fused/eager paths are byte-identical. Builds clean against the public API.

Scope note: this routes the SGD `Step` path. Adam-configured models use the separate fused
(`TryStepWithFusedOptimizer`) path, whose mixed-dtype integration is tracked separately. The plan this
calls is GPU-validated directly in Tensors (MixedPrecisionCompiledScaleGpuTests trains d512/L6/B256).


ooples:
@copilot resolve the merge conflicts in this pull request

ooples:
**Build blocker diagnosed — root cause is a Tensors release failure, not this PR's code.**

This PR references `MixedPrecisionCompiledPlan` (FP16 mixed-precision compiled plan), which lives in `AiDotNet.Tensors.Engines.Compilation`. CI fails with `CS0246: MixedPrecisionCompiledPlan could not be found` because the pinned package **AiDotNet.Tensors 0.91.12 does not contain that type** (verified by inspecting the published DLL).

The type *is* on Tensors `main` (commit `821316e7`, the FP16 `[#558]` merge), but the package that would export it — `AiDotNet.Tensors 0.92.0` — never published: the release pipeline's NuGet push aborted on the oversized native CUDA package (HTTP 413, exceeds NuGet's 250 MB limit), which under `set -e` killed the step before the managed package was pushed.

Fixed at the source in **ooples/AiDotNet.Tensors#572** (skip over-limit packages, make per-package pushes non-fatal). Once that merges and a Tensors release publishes `AiDotNet.Tensors` ≥ 0.92.0, this PR just needs its `AiDotNet.Tensors` pin bumped to that version and it will compile. No code change is required here until then.

PR #1567, opened 2026-06-11
fix(rl): WorldModelsAgent network sizing, deterministic clone, pure predict

ooples:
## Summary

Fixes the **Generated Layers N-Z** shard from the PR #1563 failing-CI list — `WorldModelsAgentTests.ActionSelection_ShouldBeFinite` and `Clone_ShouldProduceSamePolicy` (in my hand-off lane from #1562; does not touch any files #1562 owns).

Three distinct root-cause bugs, found by reproduction:

### 1. Malformed multi-GB networks → weight-streaming crash
Each VAE/RNN sub-network was constructed from a `NeuralNetworkArchitecture` with **no explicit layers**, so `NeuralNetwork.InitializeLayers` fell back to `CreateDefaultNeuralNetworkLayers` (auto-generating a hidden layer ~2× the flattened input) and the agent's `AddLayer` calls then stacked the *intended* layers on top. For a 64×64×3 = 12,288-wide observation, that default layer alone is a single **12,288×24,576 weight (~2.4 GB at FP64)** — which tripped weight-streaming's per-tensor `byte[]` cap (`int.MaxValue`) and threw `NotSupportedException` on the first forward.
**Fix:** build each network with its exact layer list passed to the architecture, so no default layers are generated.

### 2. Non-deterministic Clone
`InitializeLayers` does not wire explicitly-supplied custom layers, so their lazy weights initialized from the shared non-deterministic RNG — a clone re-initialized a *different* policy.
**Fix:** assign a deterministic per-layer `RandomSeed`; `Clone()` reproduces the networks via the constructor (they are never trained — only the controller is, via the evolution strategy) and copies the trained controller weights + rollout state directly. A serialization round-trip is **not** used for cloning because it rebuilds the layers and drops the seed pins.

### 3. Non-pure Predict
`SelectAction` advances the RNN hidden state for sequential rollout, making repeated `Predict` calls non-deterministic (surfaced `Policy_ShouldBeDeterministic` once the crash was gone).
**Fix:** `Predict` snapshots and restores the hidden state so inference is side-effect-free.

## Verification

All **7** `WorldModelsAgentTests` pass locally (`net10.0`, Release): `ActionSelection_ShouldBeFinite`, `Clone_ShouldProduceSamePolicy`, `Policy_ShouldBeDeterministic`, and 4 others.

### DIFF
--- a/src/ReinforcementLearning/Agents/WorldModelsAgent.cs
+++ b/src/ReinforcementLearning/Agents/WorldModelsAgent.cs
@@ -82,6 +82,15 @@ public class WorldModelsAgent<T> : DeepReinforcementLearningAgentBase<T>
     private int _updateCount;
     private Random _random;
 
+    // Deterministic base seeds for each sub-network's lazy weight init, so a
+    // Clone()'d agent reproduces the original's policy (the networks resolve
+    // their weights on first forward and would otherwise draw from the shared
+    // non-deterministic RNG). Distinct bases keep the three networks' inits
+    // independent.
+    private const int EncoderSeedBase = 7001;
+    private const int DecoderSeedBase = 7101;
+    private const int RNNSeedBase = 7201;
+
     /// <summary>
     /// Initializes a new instance with default settings.
     /// </summary>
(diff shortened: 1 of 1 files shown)


ooples:

Issue #1370, opened 2026-05-18
Shape Oracle: layer-side TryDeclareShape() to eliminate LoRA warmup forward (review #1368)


ooples:
## Goal — eliminate the LoRA warmup forward pass

Currently `AiModelBuilder.BuildSupervisedInternalAsync` runs a warmup
forward pass (one sample, sliced via `TrySliceFirstSampleForLoRAWarmup`)
before LoRA wrapping to materialize shapes for lazy-init layers
(LayerNormalization γ/β default, MultiHeadAttention lazy weight banks).
Without warmup, those layers report `IsShapeResolved == false` and the
LoRA wrap loop skips them.

PyTorch / HuggingFace PEFT / Microsoft LoRA avoid this entirely by
declaring shapes at module construction time (`LlamaConfig.hidden_size`).
AiDotNet's lazy-init pattern is more flexible but the perf and complexity
cost is the warmup forward pass + the `IsShapeResolved` gate.

This issue tracks the work to **beat industry standard**: layers that can
declare their shape from constructor / config args should do so
proactively, eliminating the warmup pass entirely.

---

## Design — `TryDeclareShape()` virtual on `LayerBase<T>`

```csharp
// LayerBase<T>
/// <summary>
/// Optionally declares this layer's input and output shapes WITHOUT
/// requiring a forward pass. Returns true if the layer can materialize
/// both shapes from constructor / config args alone; false if a forward
/// is required (e.g. LazyConv2d that infers input channels from the
/// first input tensor).
/// </summary>
/// <remarks>
/// Default implementation returns true iff IsShapeResolved is already
/// true (i.e. shapes were materialized at construction). Overrides on
/// LayerNormalization, MultiHeadAttention, and other lazy-init layers
/// that have enough config-time info to declare their shape should
/// materialize their internal state here and return true.
/// </remarks>
public virtual bool TryDeclareShape()
{
    return IsShapeResolved;
}
```

Layers that have enough constructor-time info to declare shape:
- `LayerNormalization<T>`: γ/β shape == feature dim, known from ctor.
- `MultiHeadAttention<T>`: Q/K/V projection shapes == [dModel, dModel/H], known from ctor.
- `DenseLayer<T>` / `FullyConnectedLayer<T>`: already eagerly resolved
  unless the input shape was a placeholder.
- `RMSNorm<T>`, `SwiGLU<T>`, `RoPE<T>`: feature dim known from ctor.

[...]

PR #1570, opened 2026-06-11
test(batchnorm): use eager ctor for init-contract assertions in deep-math tests

ooples:
## Summary

Fixes the BatchNorm cluster in the **Integration N-O** shard from the PR #1563 failing-CI list — 8 `NormalizationLayersDeepMathIntegrationTests` (in my hand-off lane from #1562; does not touch any files #1562 owns).

**Root cause:** the tests constructed the layer via the parameterless ctor `new BatchNormalizationLayer<double>()`, which is **intentionally lazy** — gamma / beta / running stats are sized on the *first forward* (a separate eager ctor `(numFeatures)` was added in #1370 for immediate allocation). The tests then read `GetGamma()` / `ParameterCount` / ran `ZeroInitGamma` **before any forward**, so they saw length/count `0`.

They were also **mutually inconsistent**: `Init_*` assumed 3 features (`gamma.Length == 3`) while `ParameterCount` assumed 5 (`== 10`). No single default could satisfy both, and making the parameterless ctor eager-default would *regress lazy BatchNorm resolution across the codebase* (every `new BatchNormalizationLayer<T>()` relying on first-forward feature inference would get a wrong fixed size).

So the production code is correct (lazy parameterless + eager `(numFeatures)`); the tests used the wrong ctor for the contract they assert.

**Fix (test-only, no assertion changed):**
- `Init_*` (gamma=1, beta=0, runningMean=0, runningVar=1) and `ZeroInitGamma` → eager ctor `(3)`. This exercises the documented eager init, which matches the **PyTorch `BatchNorm(num_features)` standard** (weight=1, bias=0, running_mean=0, running_var=1, eps default).
- `ParameterCount` → `(5)` so 2 × 5 = 10.
- `CustomEpsilon` / `CustomMomentum` → pass `epsilon: 1.0` / `momentum: 0.5` explicitly (their hand-math assumed those values, not the 1e-5 / 0.9 defaults). These forward an input, so lazy feature resolution still applies.

## Verification

All **46** `NormalizationLayersDeepMathIntegrationTests` pass locally (`net10.0`, Release) — the 8 previously-failing BatchNorm tests plus 38 that already passed (no regressions; the lazy-resolution tests are untouched).

### DIFF
--- a/tests/AiDotNet.Tests/IntegrationTests/NeuralNetworks/NormalizationLayersDeepMathIntegrationTests.cs
+++ b/tests/AiDotNet.Tests/IntegrationTests/NeuralNetworks/NormalizationLayersDeepMathIntegrationTests.cs
@@ -21,8 +21,11 @@ public class NormalizationLayersDeepMathIntegrationTests
     [Fact(Timeout = 120000)]
     public async Task BatchNorm_Init_GammaIsOnes()
     {
-        // gamma initialized to 1.0 for each feature
-        var bn = new BatchNormalizationLayer<double>();
+        // gamma initialized to 1.0 for each feature. Use the eager ctor — the
+        // parameterless ctor is lazy (gamma sized on first forward), so it has
+        // no features to inspect before a forward. Eager init mirrors PyTorch
+        // BatchNorm(num_features): weight (gamma) = 1, bias (beta) = 0.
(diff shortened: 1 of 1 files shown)


ooples:

PR #1610, opened 2026-06-15
fix(diffusion): wire predictors into weight streaming

ooples:
## Summary

- Route diffusion noise predictor forward entrypoints through the existing transparent weight streaming lifecycle.
- Add a scoped streaming guard that registers resolved lazy weights after successful forwards and cleans up abandoned or disposed streaming allocations.
- Add regression coverage for UNet, MMDiT, and UViT forced-streaming forwards plus streaming registry cleanup on predictor dispose.

## Root Cause

The foundation-scale diffusion models already had lazy layer construction and a streaming allocator path, but most concrete `NoisePredictorBase` subclasses did not engage that path before their first lazy forward. As a result, large diffusion predictors could still materialize resident weights during model-family CI instead of moving resolved weights into the bounded streaming pool.

## Notes

PR #1605 is still expected to own the clone and test dtype changes already in flight. This PR intentionally does not duplicate those fixes; it focuses on wiring the model-side streaming path.

## Validation

- `dotnet test tests\AiDotNet.Tests\AiDotNetTests.csproj --framework net10.0 --filter "FullyQualifiedName~NoisePredictorWeightStreamingTests|FullyQualifiedName~DiTWeightStreamingTests"`
- `dotnet test tests\AiDotNet.Tests\AiDotNetTests.csproj --framework net10.0 --configuration Release --no-restore --filter "FullyQualifiedName~AiDotNet.Tests.ModelFamilyTests.Diffusion.MultiDiffusionModelTests&FullyQualifiedName!~Clone_ShouldProduceIdenticalOutput" --logger "console;verbosity=minimal"`
- `git diff --check`

Known local observation: the unfiltered Release MultiDiffusion class still fails `Clone_ShouldProduceIdenticalOutput` on current `master` with `Expected 643774499 parameters, got 10342915`; that matches the #1605 clone-fix scope.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -154,10 +154,10 @@
          gated on: Dense -> LayerNorm -> GELU -> Dense now routes its between-matmul ops through the
          FP16-native path instead of the eager FP32 fallback. 0.96.0 supersedes master's 0.95.2;
          Native packages coreleased in lockstep at 0.96.0. -->
-    <PackageVersion Include="AiDotNet.Tensors" Version="0.96.1" />
-    <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.96.1" />
-    <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.96.1" />
-    <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.96.1" />
+    <PackageVersion Include="AiDotNet.Tensors" Version="0.97.2" />
+    <PackageVersion Include="AiDotNet.Native.OneDNN" Version="0.97.2" />
+    <PackageVersion Include="AiDotNet.Native.OpenBLAS" Version="0.97.2" />
+    <PackageVersion Include="AiDotNet.Native.CLBlast" Version="0.97.2" />
     <!-- Microsoft / .NET -->
     <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageVersion Include="Microsoft.Data.Sqlite" Version="10.0.9" />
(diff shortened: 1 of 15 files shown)


ooples:
## Ready — weight streaming is now the adaptive default across all models

Unblocked by the published **Tensors 0.97.2** (bumped from 0.96.1 in this PR; merged latest `master`).

With this PR, transparent weight streaming auto-engages by default in **both** model hierarchies, gated by the same 500M-parameter threshold so models below it pay nothing (no pool, no eviction):

- **`NeuralNetworkBase`** (CNNs, RNNs, transformers, etc.): already auto-engages via `TryAutoEnableWeightStreaming()`, invoked on the forward/init paths above `DefaultStreamingThresholdParams = 500M`.
- **Diffusion `NoisePredictorBase`** (separate hierarchy): this PR wires **all 11 noise predictors** (DiT, MMDiT/MMDiTX, AsymmDiT, EMMDiT, FlagDiT, FluxDoubleStream, SiT, UNet, UViT, VideoUNet) through `BeginWeightStreamingForward` → `MaybeEngageWeightStreaming` at the same 500M threshold, plus a scoped guard that registers resolved lazy weights on success and releases reservations on a failed/abandoned forward (and on dispose). `DiffusionResBlock` is a `LayerBase` sub-component, covered by its parent predictor's scope.

### Verification (net10.0, CPU)
- `NoisePredictorWeightStreamingTests` + `DiTWeightStreamingTests`: **6/6 green**.
- Core NN sanity (`FeedForwardNeuralNetworkTests` + `RecurrentNeuralNetworkTests`): **42/42 green** on 0.97.2.

Clone/dtype changes remain owned by their respective PRs (#1605); this PR is the model-side streaming wiring only.

Issue #1706, opened 2026-06-27
[CI] Umbrella tracker: HeavyTimeout models deferred from the default test gate


ooples:
## Purpose
Living tracker for models that are **correct but inherently too slow** for the default sharded test gate (`Build & SonarCloud`, per-test / 45-min budget). Per the agreed strategy, these are tagged `[Trait("Category","HeavyTimeout")]`, **excluded** from the default gate (so PRs go green), and run best-effort in the `Heavy Timeout Models (nightly)` lane. They are *deferred, not skipped* — each is worked as perf capacity allows, and **graduates back into the default shards** (drop the trait, restore a normal `[Fact(Timeout)]`) once it fits the budget.

Mechanism: **PR #1705** (`&Category!=HeavyTimeout` global filter + nightly lane).

> ⚠️ **This list is preliminary.** It's seeded from prior profiling + the failures that leaked from the (cancelled) run on PR #1689. It will be finalized against the first **completed** CI run (rerun of `28275999786` is in flight). The cancel-in-progress concurrency means most "26 failing shards" on #1689 are **cancellation artifacts, not real failures** — only the items below produced a trustworthy timeout verdict.

## Triage rules (so this issue stays scoped)
- **OOM** → *not* HeavyTimeout. Must-fix footprint (fp16-resident, lazy ctor, streaming, reclaim-teardown). Never deferred.
- **Real model bug** (assertion failures) → *not* HeavyTimeout. Fixed to green.
- **Timeout that is a regression** (was fast, now slow) → fix the perf cause, don't tag.
- **Timeout, inherently heavy at paper scale** → tag HeavyTimeout, list here.

## Confirmed timeouts (trustworthy signal from the cancelled #1689 run)
- [ ] **ModelFamily — NeuralNetworks O-R** (18 `Test execution timed out`) — incl. ViT / GEMM-bound vision
- [ ] **ModelFamily — NeuralNetworks S** (8 timeouts)
- [ ] **ModelFamily — Code/Forecast/Segment/Survival** (≥2 timeouts)
- [ ] **ModelFamily — TimeSeries/Activation/Loss** (1 timeout — identify model)
- [ ] **Integration E-G** (1 timeout — identify model)

## Preliminary heavy cast (from prior profiling — confirm per-model vs the completed run)
**NN (GEMM-bound):** VisionTransformer/ViT (~44 s/iter), VGG (~7 s/iter).
[...]

Issue #1668, opened 2026-06-23
CI: inference arena (#1661) corrupts multi-step diffusion denoise → ~10 diffusion shards


ooples:
**Source:** clean `Build & SonarCloud` run on master @ `01a4ddb4` (Tensors 0.102.12), run `28023414937`. Part of the CI-shard failure breakdown.

## Symptom
Diffusion `ModelFamily` + `Unit 03b/03d` shards fail. Where the corruption is caught it surfaces as `System.ArgumentException : Input has N channels but layer expects M channels`; where it corrupts the shared `ArrayPool` it escalates to a **native host crash with no `[FAIL]` logged** (shard dies mid-run).

## Root cause (confirmed)
The inference arena (`#1661`, `InferenceArenaSettings.Enabled`, default-ON) wraps the `DiffusionModelBase.Generate` denoise loop and `Reset()`s per step. The loop already detaches the carried latent, but **diffusion forward layers hold cross-forward cached tensors** (e.g. `DiffusionResBlock`'s pre-allocated GroupNorm output buffer; attention reshape scratch) that are first allocated *inside* the arena. `arena.Reset()` recycles that memory, so a later step's allocation **aliases a still-referenced cached buffer** → a downsample conv output emerges with a stale `[B, H*W, C]` attention layout (e.g. `[1,16,128]` instead of `[1,128,4,4]`).

Single-step `NeuralNetworkBase.Predict` is unaffected (no second step to recycle into) — ResNet/CNN pass.

## Proof
`AIDOTNET_INFERENCE_ARENA=0` turns **DDPM 38/38** and **SyncDreamer** green (both fail with it on). Confirmed locally.

## Two fix options
1. **Safe/immediate:** gate the arena OFF for the denoise loop (correctness > the perf optimization). Add `InferenceArenaSettings.DiffusionDenoiseEnabled` (default false) and gate `DiffusionModelBase.cs` (~line 676) on it.
2. **Deep/perf-preserving:** make diffusion layer caches arena-safe (GC-allocated or pinned across `Reset()`), or fix the Tensors arena so `Rent` never returns a stale-shaped recycled tensor. Higher risk; needs Tensors-side work.

## Files
`src/Diffusion/DiffusionModelBase.cs`, `src/NeuralNetworks/InferenceArenaSettings.cs`, `src/Diffusion/NoisePredictors/DiffusionResBlock.cs` (+ Tensors `TensorArena` for the deep fix).

## Acceptance
[...]

PR #1710, opened 2026-06-28
fix(diffusion): cached posEmbed corrupted by denoise-loop arena reset (non-deterministic Predict, #1706)

### LINKED ISSUE #1711: Diffusion determinism: DreamFusion / HiDream / TCD remaining after #1710
Follow-up to #1706 (umbrella) and #1710 (fixed the MMDiT/DiT `_posEmbed` arena-corruption; `StableDiffusion3.Predict_ShouldBeDeterministic` now passes).

From the real CI run (run 28293394520), these still fail `Predict_ShouldBeDeterministic` / `Clone_ShouldProduceIdenticalOutput` from **separate** sources (different predictors, not the `_posEmbed` cache #1710 fixed):

- **HiDream** (`MMDiTXNoisePredictor`): Predict + Clone fail with a *large structural* divergence (Expected -1522, Actual 105). `MMDiTXNoisePredictor : MMDiTNoisePredictor`, so it inherits the now-fixed posEmbed — the source is in MMDiTX's own overrides/components (127-line subclass). Investigate.
- **TCD** (`UNetNoisePredictor`): `Clone_ShouldProduceIdenticalOutput` fails. UNet clone-state / compiled-forward path.
- **DreamFusion** (3D, overrides `Generate`): `Clone_ShouldProduceIdenticalOutput` fails with a moderate ~1.5% gap (Expected 46.66, Actual 45.98) -> likely a clone-state copy gap.

Triage (#1706): fix the model, don't weaken the test. `CACTI`/`Riffusion` clone failures are *parallel-run flakiness* (green in isolation) -> separate concern.

Repro (AIDOTNET_DISABLE_GPU=1, net10.0):
`dotnet test tests/AiDotNet.Tests/AiDotNetTests.csproj --framework net10.0 -c Release --filter "FullyQualifiedName~HiDreamModelTests.Predict_ShouldBeDeterministic"`

ooples:
## What & why

Diffusion `Predict` was non-deterministic — `StableDiffusion3ModelTests.Predict_ShouldBeDeterministic` failed with two `Predict` calls on the same model+input returning wildly different output (9.78 vs 6.00 — not a precision issue). Root-caused from #1706's real CI failure list.

## Root cause

`MMDiTNoisePredictor` / `DiTNoisePredictor` cache the sinusoidal position embedding `_posEmbed` (computed once per patch count, reused on every forward) but allocated it via **`TensorAllocator.Rent`** — arena Tier-0 scratch. `DiffusionModelBase.Generate` runs the denoise loop inside a `TensorArena` and calls `Reset()` **every step** to recycle per-step scratch. So the cached `_posEmbed` aliased recycled scratch and was **corrupted between steps**, adding garbage to every image token → the noise prediction (and the final output) differed every `Predict`, and clone-vs-original diverged identically.

Same class as #1668 (arena scratch aliasing live state), but for a **lazily-cached** tensor that `InferenceMode`'s `ShouldCacheForBackward` gate doesn't cover.

## Fix

Allocate the long-lived `_posEmbed` cache as a plain GC `new Tensor<T>` (the sinusoidal fill is cheap and runs once), so it survives the per-step arena `Reset()`. Per-step denoise scratch is untouched.

## Verification (AIDOTNET_DISABLE_GPU=1, net10.0, in isolation)
- `StableDiffusion3.Predict_ShouldBeDeterministic` — **now passes** (was the confirmed failure).
- `CACTI.Clone_ShouldProduceIdenticalOutput` — green.
- Build succeeded.

## Scope
This now resolves **all** of the #1706 diffusion-determinism failures, not just the posEmbed class:

- **posEmbed arena-corruption** (MMDiT/DiT) — the original fix above.
- **HiDream (MMDiTX), TCD (UNet), DreamFusion (3D)** — fixed in a follow-up commit on this branch (no longer deferred). They had **two further root causes**, both lazy-weight materialization timing under the denoise-loop arena:
  1. **Predict non-determinism (HiDream)** — a freshly-constructed predictor materializes its `LazyDense` weights on its *first forward*; when that ran inside the transient per-`Generate` arena, the `RentPinned` weights were freed on the arena's dispose, so the next `Predict` read recycled memory (arena-on diverged ~5e5; arena-off exact). `DiffusionModelBase.Generate` now does a one-time warmup forward **outside** the arena so lazy weights land on the GC heap.
  2. **Clone non-fidelity (all three)** — clones copied parameters onto *unmaterialized* lazy layers, which then re-RNG-initialized on the clone's first forward and discarded the copied weights. Fixed by materialize-then-copy: extracted the base predictor's probe+copy into `ProbeMaterializeAndCopyInto` (used by MMDiTX); `TCDModel`/`DreamFusionModel` now delegate to their sub-models' own `Clone()` (which materialize then copy).

Verified (CPU): `HiDream/TCD/DreamFusion.Clone_ShouldProduceIdenticalOutput` + `HiDream.Predict_ShouldBeDeterministic` → **4 passed / 0 failed**; isolation probe maxDiff now 0 across all arena-on/clone cases (was 5e5 / 7e2 / 3.8 / 2.8e1). Builds clean on net10.0 + net471.

Closes #1711.
Refs #1706.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where cached positional embeddings could be reused incorrectly across diffusion steps, leading to unstable or corrupted results.
  * Improved consistency of denoising predictions by keeping positional embedding data available for the full processing flow.

### DIFF
--- a/src/Diffusion/DiffusionModelBase.cs
+++ b/src/Diffusion/DiffusionModelBase.cs
@@ -174,6 +174,11 @@ public abstract class DiffusionModelBase<T> : IDiffusionModel<T>, IConfigurableM
     /// </summary>
     private Tensor<T>[]? _cachedTrainableParameters;
 
+    // #1706: set once the noise predictor's lazy weights have been materialized OUTSIDE the
+    // transient per-Generate denoise arena (see the warmup in Generate). Stops the first Generate
+    // from allocating model-lifetime weights into that arena, which frees them on dispose.
+    private bool _lazyWeightsWarmed;
+
     // ── Copy-on-write weight sharing (cheap Clone of large models) ───────────────────────────────
     //
     // Clone() shares the parent's weight TENSORS by reference instead of deep-copying every
@@ -664,6 +669,28 @@ public abstract class DiffusionModelBase<T> : IDiffusionModel<T>, IConfigurableM
 
         // Pre-allocate reusable noise prediction vector to avoid per-step allocation
         var noisePredVec = new Vector<T>(sample.Length);
+
+        // #1706 (HiDream/MMDiTX; #1710 follow-up): a freshly-constructed noise predictor
+        // materializes its lazy weights (LazyDense / conv / embedding) on its FIRST forward, via
+        // the forward path. If that first forward runs inside the transient per-Generate arena
+        // created below, those model-lifetime weights are RentPinned into it and freed when the
+        // arena disposes at the end of THIS call — so the NEXT Generate reads recycled memory and
+        // Predict becomes non-deterministic (confirmed: arena-on differs by ~5e5, arena-off is
+        // exact, and a single warmup outside the arena restores determinism). Materialize them
+        // once here with a single warmup forward OUTSIDE the arena so they land on the GC heap
+        // (or an outer, longer-lived arena) and survive. One forward, one time per model; the
+        // denoise loop's per-step arena recycling below is unchanged.
+        if (!_lazyWeightsWarmed)
+        {
+            int warmupTimestep = 0;
+            foreach (var t in _scheduler.Timesteps) { warmupTimestep = t; break; }
+            sample.AsSpan().CopyTo(sampleTensor.AsWritableSpan());
+            using (InferenceMode.Enter())
+            {
+                PredictNoiseStep(sampleTensor, warmupTimestep);
+            }
+            _lazyWeightsWarmed = true;
+        }
 
         // Forward caching allocator (Tensors #661 consumer wiring, second boundary
         // after NeuralNetworkBase.Predict): a NoisePredictorBase forward is NOT a
--- a/src/Diffusion/FastGeneration/Flux2SchnellModel.cs
+++ b/src/Diffusion/FastGeneration/Flux2SchnellModel.cs
@@ -148,10 +148,13 @@ public class Flux2SchnellModel<T> : LatentDiffusionModelBase<T>
     /// <inheritdoc />
     public override IDiffusionModel<T> Clone()
     {
-        var clone = new Flux2SchnellModel<T>(conditioner: _conditioner, seed: RandomGenerator.Next());
-        // #1624: O(1)-until-write copy-on-write parameter share (avoids the full-model flatten copy that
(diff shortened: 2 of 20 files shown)


ooples:
Pushed additive commits on top of `9b02619ac` (the trio fix) — kept your MMDiT/MMDiT-X `ProbeMaterializeAndCopyInto` approach, renamed my callers to match it. Added:

- **DenseLayer GC-weights** (`0c8bef810`) — lazy dense weights must be GC-owned, not arena-recyclable (#1711).
- **DiT-predictor clone sweep** — applied the probe-forward-then-copy clone semantics across the rest of the DiT family so their `Clone_ShouldProduceIdenticalOutput` is correct:
  - `AsymmDiT`/`EMMDiT`/`FluxDoubleStream` predictors now call the base MMDiT `ProbeMaterializeAndCopyInto`.
  - Extracted `DiTNoisePredictor.ProbeMaterializeAndCopyInto` and reused it in `SiT`.
  - `FlagDiT` got a probe-forward + per-layer copy clone (lazy GQA).
  - Converted 8 pattern-B models whose model-level `SetParameters(GetParameters())` clone bypassed the predictor's `Clone` to delegate to `predictor.Clone()`/`vae.Clone()`: SiDDiT, MAR, PixArtDeltaLCM, MotionDiffusion, SD3Flash, SD3Turbo, SenseFlow, Flux2Schnell.

**Verified against released AiDotNet.Tensors 0.104.5** (PR CI baseline): TCD + HiDream + SiDDiT `Clone_ShouldProduceIdenticalOutput` pass 3/3; build green; no regression on NeuralTuringMachine/DifferentiableNeuralComputer (42/42) or ConvolutionalNeuralNetwork (21/21).

**Companion engine fix:** ooples/AiDotNet.Tensors#708 (`InvalidatePersistentTensor` now invalidates the per-array derived-weight caches) — the root-cause fix for the same class; not required for this PR's CI (the AiDotNet-side trio fix already passes against released Tensors), but lands the general correctness fix.

**Follow-ups (not in this PR):** foundation-scale `Flux2Schnell`/`SANA` clones are now *correct but exceed the 120s test timeout* (HeavyTimeout candidates); ControlNet diffusion models (extra control encoder) left untouched.

PR #1769, opened 2026-07-02
perf(optimizers): SIMD-vectorize the fused Adam/AMSGrad Step kernel (#1757)

### LINKED ISSUE #1757: perf(optimizers): SIMD-vectorize the Adam Step kernel inner loop
## Context
While making the diffusion training step paper-faithful + configurable (#1748), we found the shared Adam optimizer's fused tape-step kernel (`AdamOptimizer.Step(TapeStepContext)`) already avoids per-step allocations (in-place `MemoryMarshal`/`Span` update over the raw `double`/`float` backing arrays — it replaced an engine-op chain that allocated ~15+ intermediate tensors/step). **But its inner loop is scalar** (`Unsafe.Add` one element at a time), not hardware-SIMD.

We already have SIMD primitives in the `AiDotNet.Tensors` engine (`SimdKernels`, used by `Engine.*`). The Adam kernel should use the same approach.

## Task
Vectorize the `double`/`float` fast-path inner loops in `AdamOptimizer.Step` (and likely `AdamWOptimizer`) with `System.Numerics.Vector<double>/<float>` (or the existing Tensors SIMD primitives): compute `m = β1·m + (1-β1)·g`, `v = β2·v + (1-β2)·g²`, `m̂/v̂` bias correction, and `param -= lr·m̂/(√v̂+ε)` a vector-width at a time, with a scalar tail and the generic-`T` fallback preserved. Include the AMSGrad `Vector.Max` branch.

## Value / caveat
Benefits every optimizer-bound training path. Note: end-to-end training of large models is frequently forward/GEMM-bound (e.g. #1748's Step1XEdit is DiT-XL-scale SiT GEMM FLOPs — the optimizer is a tiny fraction), so the practical win is on optimizer-bound workloads, not universally.

## Acceptance
- Bit-reasonable numerical parity with the current scalar kernel (existing optimizer unit tests pass).
- Before/after **median-of-N** timing on a large flat parameter vector showing the speedup.
- Broad regression check across optimizer consumers (NN + diffusion training smoke tests).

Blast radius: the optimizer is used by every model — treat as its own focused PR with proper timing + regression testing.

ooples:
## Summary
The tape-step Adam kernel (`AdamOptimizer.Step(TapeStepContext)`) already avoided per-step allocations (in-place update over the raw backing arrays), but its **inner loop was scalar** (`Unsafe.Add` one element at a time). This extracts the moment update into a shared `AdamMomentSimd.Step` and vectorizes it with `System.Numerics.Vector<T>` — `mNew`/`vNew`, bias-correction divide, sqrt, param update, and the AMSGrad running-max all run `Vector<T>.Count` lanes at a time, with a scalar tail for the remainder.

All four call sites (double/float × live-array/buffer-aliased) now route through the one kernel.

## Bit-identical — by construction and by test
This is a *faster kernel that changes no numbers*:
- **fp64** uses `Vector.SquareRoot` (== scalar `Math.Sqrt`) and separate mul/add (no FMA contraction), so every lane matches the scalar loop bit-for-bit.
- **fp32** widens to fp64 for the sqrt then narrows — matching the scalar path's `(float)Math.Sqrt(...)` **exactly**, not a ~1-ULP single-precision sqrt.
- **AMSGrad** max uses `ConditionalSelect(GreaterThan(...))` to match the scalar `>` ternary (incl. NaN behaviour).

## Tests
- New `AdamMomentSimdParityTests`: asserts **bit-identical** output vs an independent scalar reference across **9 fp64 + 8 fp32 sizes** (including sub-vector-width tails: 1, 3, 7, 15, 31, …) × {Adam, AMSGrad} — 17 theories, all green.
- Existing `FusedOptimizerParityTests` / `AdamOptimizerAnomalyGuardTests` / `FusedActivationParityTests` unchanged (29 green).
- Both net10.0 and net471 targets build (fp32 sqrt uses widen/narrow, not `MathF`, for net471 safety).

Closes #1757


ooples:
Closing in favor of the correct layering. The SIMD Adam step should not be hand-rolled in the consumer with `System.Numerics.Vector` — it belongs in **AiDotNet.Tensors** alongside the existing `FusedOptimizer.AdamUpdateSimd` / `SimdVector`.

Replaced by **ooples/AiDotNet.Tensors#721**: a shared `AdamMomentKernels.AdamStep` (AVX2+FMA, float+double, AMSGrad) that is bit-identical SIMD↔scalar (17 parity theories). Once that Tensors release lands, the eager `AdamOptimizer` tape-step will call it (via `InternalsVisibleTo "AiDotNet"`) in a small follow-up, closing #1757.

This branch has been reverted to master (hand-rolled bandaid removed). Also for the record, the two issues on this PR were: a net471 test build break (`BitConverter.SingleToInt32Bits` is not on net471) and a commit-subject casing lint fail — both avoided in the Tensors PR.

PR #1805, opened 2026-07-06
feat(facade): pluggable credit-assignment rules (Feedback Alignment / DFA / Sign-Symmetric)

ooples:
## What

Adds a public, extensible **credit-assignment ("learning rule") system** so users can train through `AiModelBuilder.BuildAsync()` with standard back-propagation OR published non-backprop rules — selected entirely through the facade, with the default path unchanged.

Motivation: a research effort (HarmonicEngine B3) needs to prove zero-/non-backprop credit assignment scales like backprop, using the same forward pass, optimizer, batching and scheduler. This gives a clean seam and a verified reference implementation of the feedback-alignment family, plus an interface a private/novel rule can implement without being exposed here.

## Public API

- **`ICreditRule<T>`** + **`ICreditAssignmentContext<T>`** + **`ICreditLayer<T>`** (`src/Interfaces/ICreditRule.cs`): the general contract *"given forward activations + output error, produce per-layer parameter updates."* Not hardcoded to any one rule.
- **`CreditRule`** enum `{ Backprop, FeedbackAlignment, DirectFeedbackAlignment, SignSymmetric }` (`src/Enums/CreditRule.cs`).
- **`AiModelBuilder.ConfigureCreditRule(CreditRule rule, int? seed = null)`** and **`ConfigureCreditRule(ICreditRule<T>? rule, int? seed = null)`** on `IAiModelBuilder`.

## Seam

`NeuralNetworkBase<T>.ComputeGradients` — the single per-parameter-gradient producer that **both** facade training paths funnel through (`AdamOptimizer.Optimize → CalculateGradient → ComputeGradients`, and the direct path) — branches to the configured rule and returns the flat gradient vector in `GetParameters()` order, so the optimizer/batching/scheduler are untouched. When no rule is configured (or `CreditRule.Backprop`, which maps to a null rule), the reverse-mode tape path is **byte-for-byte unchanged**. The rule is wired onto the model from the builder via `SetCreditRule`.

## Rules shipped (feedback-alignment family, dense feed-forward stacks)

| Rule | Feedback path |
|------|---------------|
| Backprop (reference) | true transpose weights |
| FeedbackAlignment | fixed random matrix per layer (Lillicrap 2016) |
| DirectFeedbackAlignment | global output error projected to each layer via fixed random matrix (Nøkland 2016) |
| SignSymmetric | sign of the transpose weights (Liao 2016) |

Scope: a stack of `FullyConnectedLayer<T>` with a matched output loss (MSE+linear or cross-entropy+softmax) — the domain in which the feedback-alignment literature is defined. Non-supported architectures raise a clear `NotSupportedException` rather than silently misbehaving.

## Verification (`tests/.../CreditRuleFacadeTrainingTests`, 6 tests, all green)

Measured the **correct** way — held-out TOP-1 accuracy (not the streamed optimizer loss; not double-softmaxed perplexity):

- **End-to-end via `BuildAsync`** on a learnable 3-class Gaussian-blob task: FeedbackAlignment / DirectFeedbackAlignment / SignSymmetric each reach **≥ 0.65 held-out accuracy** (chance = 0.333) with **> +0.10** over the untrained baseline.
- **Fixed-net gradient check**: the credit-rule engine's own back-prop reproduces the tape gradient **direction (cosine > 0.99)**, and FA/DFA/Sign-Symmetric are **positively aligned** with true back-prop (they self-align, as the literature predicts).
- **Default path unchanged**: all 8 existing `FacadeTrainingCallbackTests` pass.

## Deferred (intentionally, not stubbed)

`e-prop`, `Forward-Forward`, `Target-Prop` are not included — the public `ICreditRule<T>` is general enough to add them later or implement them privately. No dead code / stubs.

### DIFF
--- a/src/AiModelBuilder.BuildPipeline.cs
+++ b/src/AiModelBuilder.BuildPipeline.cs
@@ -1208,6 +1208,15 @@ public partial class AiModelBuilder<T, TInput, TOutput>
         // which caused race conditions when multiple models were built concurrently.
         ConfigureDocumentTransformers(_model);
 
+        // Wire the pluggable credit-assignment rule (ConfigureCreditRule) onto the neural network. When set,
+        // NeuralNetworkBase.ComputeGradients routes the error to each layer through the rule (Feedback
+        // Alignment / Direct Feedback Alignment / Sign-Symmetric / a custom ICreditRule) instead of standard
+        // back-propagation. Null leaves the default path byte-for-byte unchanged.
+        if (_creditRule is not null && _model is NeuralNetworks.NeuralNetworkBase<T> creditRuleNet)
+        {
+            creditRuleNet.SetCreditRule(_creditRule, _creditRuleSeed);
+        }
+
         // Use defaults for the optimizer if not set. ConfigureRegularization
         // and ConfigureDistributedTraining both require gradient semantics:
         // regularization is applied in GradientBasedOptimizerBase, and DDP
--- a/src/AiModelBuilder.Configure.cs
+++ b/src/AiModelBuilder.Configure.cs
@@ -332,6 +332,50 @@ public partial class AiModelBuilder<T, TInput, TOutput>
         return this;
     }
 
+    /// <summary>
+    /// Selects the <b>credit-assignment (learning) rule</b> used to produce per-parameter updates during
+    /// neural-network training. The default is <see cref="CreditRule.Backprop"/> (standard reverse-mode
+    /// back-propagation); alternative published rules — Feedback Alignment, Direct Feedback Alignment,
+    /// Sign-Symmetric — replace how the error signal is routed back to each layer while reusing the same
+    /// forward pass, optimizer, batching and scheduler.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// <b>For Beginners:</b> when the network is wrong, it must decide how much each weight was to blame.
+    /// Back-prop does this exactly (through the transpose of every weight); the feedback-alignment rules use a
+    /// fixed random (or sign-only) shortcut instead — and still learn. Use this to experiment with those
+    /// alternatives without changing anything else about training.
+    /// </para>
+    /// <para>
+    /// Selecting <see cref="CreditRule.Backprop"/> is identical to not calling this method at all (the default
+    /// path is left byte-for-byte unchanged). The non-backprop rules currently support a dense feed-forward
+    /// stack (<c>FullyConnectedLayer</c> layers) with a matched output loss.
+    /// </para>
+    /// </remarks>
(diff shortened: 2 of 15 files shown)


ooples:
## Update: credit rules now support Transformers (not just dense stacks)

Reworked the engine so the non-backprop rules apply to **any layer type** — multi-head attention, feed-forward, LayerNorm, embedding — addressing the review that throwing `NotSupportedException` for transformers was a punt.

### Unified tape-VJP mechanism
The forward pass runs under the gradient tape, capturing each trainable layer's output. A rule supplies a per-layer **teaching signal** at that output; the engine converts it into that layer's parameter gradients via a **local vector-Jacobian product** (one backward step through only that layer, which automatically supplies the layer's activation Jacobian). The output layer is trained with the exact loss gradient. Gradients are flattened by keying on parameter-tensor references, so composite Transformer blocks work without special-casing.

This is exactly why **Direct Feedback Alignment scales to attention** (Launay et al. 2020): DFA needs no per-layer backward chain — each layer's teaching signal is a fixed random projection of the global output error.

### Rules
| Rule | Scope |
|------|-------|
| DirectFeedbackAlignment | **General** — Transformers + dense |
| FeedbackAlignment | General (sequential fixed random feedback) |
| SignSymmetric | Dense only — throws a clear `NotSupportedException` on attention/norm blocks (no single chaining weight); DFA is the transformer path |
| Backprop | `IsExactBackprop` → default reverse-mode path, byte-for-byte unchanged |

### API — all three styles coexist (per request)
- `ConfigureCreditRule(CreditRule.DirectFeedbackAlignment, seed: 42)` — enum
- `ConfigureCreditRule(CreditRules.DirectFeedbackAlignment<float>(seed: 42))` — new **`CreditRules`** static class of configurable instances
- `ConfigureCreditRule(myCustomRule)` — `ICreditRule<T>`

### Verification (8/8 green, CPU engine via `AIDOTNET_DISABLE_GPU=1`)
- **`ConfigureCreditRule(DirectFeedbackAlignment)` trains an actual `Transformer<float>`** (d_model 48 / 2 encoder layers / 4 heads, SequenceClassification) through `BuildAsync` to held-out **top-1 accuracy ≥ 0.55** (chance 0.333), **> +0.10** over the untrained baseline — measured by accuracy, not the streamed loss.
- DFA / FA / SignSymmetric each train a dense MLP to ≥ 0.60 held-out accuracy.
- DFA / FA / SignSymmetric gradients positively align with true back-prop on a fixed net.
- SignSymmetric throws `NotSupportedException` on a Transformer; DFA does not.
- Default path unchanged: all 8 existing `FacadeTrainingCallbackTests` pass.

CPU engine used deliberately to isolate credit-rule correctness from the separate CUDA transformer-training issue another agent is fixing.

Issue #1804, opened 2026-07-06
Deep time-series models train single-threaded / underutilize GPU (unbatched per-op tape)


ooples:
## Summary
The deep time-series models (N-BEATS, DeepAR, Informer, TFT, N-HiTS, Autoformer) train **single-threaded** — 1 of 128 CPU cores, ~0–10% GPU — because each training step is a long tape of *small* per-op tensor operations rather than a few large resident ops. Classical models (Matrix.Multiply GEMM) and raw engine GEMM parallelize fine, so this is specific to the deep-model training path.

## Measured (AiDotNet 0.229 master + Tensors 0.110)
- N-BEATS train, default **BatchSize=32**: **1.0 core**, 0% GPU.
- N-BEATS train, **BatchSize=512**: **1.0 core**, ~10% GPU (bigger ops reach the GPU but the tape still runs serially).

## Root cause
`NBEATSModel.TrainCore` (and the sibling deep models) run a `GradientTape` whose forward+backward is a chain of small ops (elementwise `TensorSubtract`/`TensorAdd`, per-block GEMMs at hidden≈512). With the default batch these GEMMs are below the engine's parallel threshold; even at large batch the *serial per-op tape* + host round-trips keep the GPU idle between ops.

## Fixes (pairs with AiDotNet.Tensors#728)
1. **Raise the deep-model default batch** (currently 32) so per-step GEMMs are large — a quick partial win.
2. **Restructure the train step** to fewer, larger, GPU-resident operations (batch the whole stack forward/backward; keep tensors resident; let Tensors fuse/graph-capture the step) so the models actually saturate CPU cores / the GPU.

## Repro
`NBEATSModel<float>` (hidden 512, 3×3 blocks, BatchSize 32 vs 512) on a 20k-pt series; sample process CPU-seconds + nvidia-smi during Train.

ooples:
### Quantified: training is ~275x per-op overhead, not compute

Timed N-BEATS<float> (n=8000, lookback 96, hidden 256, 3x3 blocks, batch 64, 20 epochs, forced CpuEngine): **~275s**.

[...]

PR #1809, opened 2026-07-06
perf: activate TensorArena for training across all model paths (DRAFT)

ooples:
**Draft — WIP.** Core mechanism validated on N-BEATS; regression/invariant test suite being finalized.

## What
Open a `TensorArena` around every training loop so per-op forward/backward scratch is recycled instead of GC-allocated. Pairs with the Tensors-side per-step reset (ooples/AiDotNet.Tensors#734 — **required for the reuse/reset behavior**).

- `TimeSeriesModelBase.Train` wraps `TrainCore` in an arena → every TS model inherits it transparently (mirrors `NeuralNetworkBase`/`DiffusionModelBase`).
- **`Optimize()` path widened:** the real `optimizer.Optimize` invocation sites (`AiModelBuilder.BuildPipeline` ×4, `.Internals`, both cross-validators) run inside an arena — covering all 39 optimizers transitively (regression / classical / facade training).

Weights + Adam moments are plain-heap tensors → they survive the per-step reset; only Engine-op scratch is arena-backed.

## Evidence (N-BEATS)
- Correctness: forecast + MAE **bit-identical** to no-arena.
- Allocation: 11,714.9 MB → **1,707.9 MB (85.4% cut)**.

## Still to land before un-drafting
- Model-equivalence (arena-on ≡ arena-off) + invariant tests (in `tests/.../IntegrationTests/Arena/`) — being finalized.
- A possible follow-up: global tape-owned arena for any custom loop not going through these bases.

Contributes to #1804.

### DIFF
--- a/src/AiModelBuilder.BuildPipeline.cs
+++ b/src/AiModelBuilder.BuildPipeline.cs
@@ -22,6 +22,209 @@ namespace AiDotNet;
 public partial class AiModelBuilder<T, TInput, TOutput>
 {
     /// <summary>
+    /// Adapts the optimizer's per-epoch seam (and the direct-training path's single pass) to the
+    /// facade's shared per-epoch callback/monitor logic, tracking the abort state for the result
+    /// surface. Allocated only when a monitor or training callback is configured, so the default
+    /// training path never constructs one.
+    /// </summary>
+    private sealed class EpochProgressBridge
+    {
+        private readonly AiModelBuilder<T, TInput, TOutput> _owner;
+        private readonly string? _monitorSessionId;
+        private readonly CancellationToken _cancellationToken;
+        private readonly int _totalEpochs;
+        private readonly DateTime _start;
+
+        /// <summary>Zero-based index of the last epoch observed.</summary>
+        public int LastEpoch { get; private set; }
+
+        /// <summary>Loss reported for the last epoch observed.</summary>
+        public T LastLoss { get; private set; }
+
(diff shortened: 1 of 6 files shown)


ooples:
## Validation complete — arena is bit-equivalent across architectures ✅

Multi-model equivalence via the coreprobe harness: each model trained TWICE with a fixed seed — arena OFF (`TensorPool.ForceFreshAllocations=true`) vs arena ON — predictions compared element-by-element.

| Model (path) | max abs diff (ON vs OFF) | train alloc OFF→ON | Result |
|---|---|---|---|
| N-BEATS (TimeSeries tape) | **0.0** (bit-exact) | 107.58 → 47.62 MB (−55.7%) | PASS |
| MLP (eager NeuralNetworkBase) | **0.0** (bit-exact) | 44.01 → 42.26 MB (−4.0%) | PASS |
| Facade `Optimize()` (the new widening) | **0.0** (bit-exact) | 21.69 → 2.03 MB (−90.7%) | PASS |

Plus: Tensors reshape regression unit test **5/5 green** (net10.0 + net471); N-BEATS full-run bit-identical with **85.4%** allocation cut.

**Honest caveat (not inflating numbers):** the eager `NeuralNetwork.Train` path only drops 4% — its scratch is dominated by non-arena Matrix allocations, so the arena barely reduces its total there (correctness still holds bit-exact). The large wins are on the tape and `Optimize()` paths, which is exactly where the reuse engages.

Un-drafting for review.

PR #1816, opened 2026-07-07
perf: cache Adam per-param backing arrays across steps (N-BEATS min-of-7 -> 15.7s, beats PyTorch)

ooples:
Re-profile showed `AdamOptimizer.Step`'s float fast path re-resolves each param's m/v every step (2 ConcurrentDictionary lookups + shape `SequenceEqual` + 4 `GetLiveBackingArrayOrNull` calls, x~90 params/step) — pure steady-state overhead since param refs are stable. Cache {param -> m/v backing arrays + length} keyed by identity, resolve once. **Bit-exact** (same arithmetic).

Combined with the GEMM parallel-threshold fix (ooples/AiDotNet.Tensors#740), N-BEATS CPU-train **min-of-7: 28.9s -> 15.7s**, under eager PyTorch's 16.6s reference on this 128-core box; forecast + MAE identical. Contributes to #1804.

### DIFF
--- a/src/Optimizers/AdamOptimizer.cs
+++ b/src/Optimizers/AdamOptimizer.cs
@@ -189,6 +189,7 @@ public class AdamOptimizer<T, TInput, TOutput> : GradientBasedOptimizerBase<T, T
         _tapeM.Clear();
         _tapeV.Clear();
         _tapeVMax.Clear();
+        _fp32StepCache.Clear();
         _tapeStep = 0;
 
         // Initialize parameters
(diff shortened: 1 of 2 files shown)


ooples:
**Correction to the perf claim in this PR body.** The "28.9s -> 15.7s, beats PyTorch" figure was thermally confounded (the box throttles ~2x; the 28.9s reference was a hot state). The **honest, thermal-controlled result** for the changes in this PR (Adam per-param cache + column-shaped biases): **bit-exact** (forecast[0]=47.395111, MAE=0.023372) and a real **~4-5% wall + -1.4% allocation** (thermal-independent) win vs the same-session baseline. The cool baseline is already ~15.5-17.2s ≈ PyTorch's 16.6s; these changes give a genuine small edge on top, purely by removing per-step overhead (no numeric change). The separate GEMM-threshold PR (Tensors #740) that the body credited was measured to be a REGRESSION and has been closed. This PR stands on its own bit-exact merits.

PR #1896, opened 2026-07-20
feat(rag): productionize the RAG pipeline (real LLM, real vector stores, async, quantization, eval, facade wiring)

ooples:
## RAG productionization

Turns the RAG subsystem from a broad-but-partly-stubbed surface into a functional, production-oriented stack, and makes all of it reachable through the AiModelBuilder facade (code + YAML). Motivated by a competitive-parity audit (LangChain / LlamaIndex / Haystack / Semantic Kernel; Pinecone / Weaviate / Qdrant / Milvus / Chroma / pgvector / FAISS).

### What changed (18 workstreams, one commit each)
- **Real LLM generation** — ChatClientGenerator bridges the production IChatClient connectors (OpenAI/Azure/Anthropic/Cohere/Gemini/Mistral/Ollama) with streaming; de-stubbed HyDE, LLM/Cohere rerankers, LLM context compression, LLM query expansion, multi-query, and GraphRAG entity/relation/claim extraction + community reports to actually call it.
- **Real vector stores (7/8)** — Qdrant, Pinecone, Weaviate, Milvus, Azure AI Search (HTTP), pgvector (AiDotNet.Storage.Postgres), RedisVL (AiDotNet.Storage.Redis); the in-memory simulations were deleted from core.
- **Performance** — scalar/binary/product quantization + ADC, content-hash embedding cache, real ONNX cross-encoder reranker.
- **Quality** — RRF hybrid fusion + ensemble/routing/self-query retrievers; RAGAS + IR metrics (nDCG/MRR/MAP/hit-rate); embeddings now fail loudly instead of returning fake vectors; token-aware + real embedding-based semantic chunking.
- **Async** — IRetriever/IDocumentStore/IReranker/IContextCompressor/IGenerator are async + cancellation-aware; I/O-bound stores/generators are truly async.
- **Facade** — ConfigureVectorStore/ConfigureVectorIndex + chunking/compression wiring (interface AND concrete), the documentStore-dropped-without-a-KG bug fixed, ConfigureRAG, and a YAML composite section wiring every sub-component. Central RagPipeline<T> orchestrator (ingest/query, async, tenant isolation).

### Verification
- Per task: builds green on net471/net8/net10; ~230 new unit tests. Live-service tests are gated ([Trait("Category","Integration")] + env vars) and skip in CI.
- A full RAG-suite run on the integrated branch is being completed; any failures found will be fixed before merge.

### NOT done — gaps vs. "exceed every competitor" (follow-ups, tracked not glossed)
[...]

PR #1918, opened 2026-07-21
fix(rag): harden frontier retrieval paths

ooples:
## Summary

- require real token embeddings for ColBERT and add an adapter for existing embedding models
- harden tenant isolation by namespacing physical IDs while preserving caller-visible IDs
- correct cosine ANN/PQ metric consistency and add recall regression coverage
- add local end-to-end RAG coverage plus cancellation and isolation tests

## Why

This is the production-readiness follow-up to #1896. It removes the synthetic ColBERT fallback, prevents cross-tenant ID collisions in shared stores, and fixes cosine PQ/IVFPQ training so centroid assignment matches Euclidean mean updates.

## Validation

- complete RAG namespace suite: 2,011 passed, 0 failed, 9 skipped (live-service tests requiring credentials/infrastructure)
- focused changed-area suite: 47 passed, 0 failed
- git diff --check: clean
- full net10 repository suite: exceeded the 15-minute local timeout without emitting an early failure or aggregate result

## Notes

The existing SQLite vulnerability and Pomelo/EF Core version warnings are repository-level dependency issues and are not introduced by this patch.


ooples:
Closing this stacked draft because the completed RAG frontier commits have now been applied directly to feat/rag-productionization for PR #1896.

PR #1967, opened 2026-08-07
refactor(1789): tests + Models (02/18)

ooples:
Slice 2 of 18 splitting #1789 into review-sized PRs.

**Scope:** tests + Models

**Files:** 96

## Why an integration branch

## Integrity

Related: #1789


src/Models/Options/ChronosFinanceOptions.cs
```diff
@@ -125,6 +125,23 @@ public ChronosFinanceOptions(ChronosFinanceOptions<T> other)
...
+    /// historical context — and then bins uniformly within <c>[-bound * s, +bound * s]</c>.
+    /// Ansari et al. 2024 (arXiv:2403.07815) use a bound of 15, which is the default here.</para>
+    /// <para><b>For Beginners:</b> This controls how far above and below the typical size of
+    /// your data the model is able to express a forecast. The scale <c>s</c> is the average
+    /// magnitude of your recent history, so a bound of 15 means the model can represent values
+    /// up to fifteen times that average in either direction. Raise it if your series has rare
+    /// but very large spikes; lower it to spend the available bins on a narrower band and get
+    /// finer resolution where the data actually lives.</para>
+    /// </remarks>
+    public double QuantizationBound { get; set; } = 15.0;
```
ooples, line 144: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'ChronosFinanceOptions' does not copy property 'QuantizationBound'**

src/Models/Options/DiffusionModelOptions.cs
```diff
@@ -105,6 +106,32 @@ public DiffusionModelOptions(DiffusionModelOptions<T> other)
...
+    /// This exists because the count was previously a <c>private const int LATENT_CHANNELS = 4</c>
+    /// baked into each model, invisible and unreachable from the facade. Four is the Stable-Diffusion
+    /// value and a fine DEFAULT, but SDXL-class and audio latent models legitimately differ, and a
+    /// private const gives callers no way to say so.
+    /// </para>
+    /// <para><b>For Beginners:</b> A latent diffusion model does not work on pixels directly. It
+    /// first compresses the image into a smaller "latent" form; this is how many channels that
+    /// compressed form has. Leave it unset unless you know you need something else.</para>
+    /// </remarks>
+    public int? LatentChannels { get; set; }
```
ooples, line 134: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'DiffusionModelOptions' does not copy property 'LatentChannels'**

src/Models/Options/RWKVForecastingOptions.cs
```diff
@@ -85,4 +90,69 @@ public RWKVForecastingOptions(RWKVForecastingOptions<T> other)
...
+
+    /// <summary>
+    /// Gets or sets the variance floor used by RevIN's per-instance standard deviation,
+    /// std = sqrt(var + eps). Default: 1e-5, the value Kim et al. 2022 use.
+    /// </summary>
+    /// <remarks>
+    /// This is what keeps a constant (zero-variance) series from dividing by zero. It is exposed
+    /// because a series whose natural scale is far from unity may need a different floor.
+    /// </remarks>
+    public double RevInEpsilon { get; set; } = 1e-5;
```
ooples, line 133: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'RWKVForecastingOptions' does not copy property 'RevInEpsilon'**

ooples (commented):
**Actionable comments posted: 4**

src/Models/Options/RWKVForecastingOptions.cs
```diff
@@ -85,4 +90,69 @@ public RWKVForecastingOptions(RWKVForecastingOptions<T> other)
...
+    /// <summary>
+    /// Gets or sets whether reversible instance normalization (RevIN) is applied around the forecast.
+    /// Default: true, which is the published behaviour.
+    /// </summary>
+    /// <remarks>
+    /// RevIN is an optional wrapper in Kim et al. 2022, not part of the RWKV architecture itself, so it
+    /// is switchable. Turning it off makes the model consume the raw series and forecast on the raw
+    /// scale.
+    /// </remarks>
+    public bool UseReversibleNormalization { get; set; } = true;
```
ooples, line 144: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'RWKVForecastingOptions' does not copy property 'UseReversibleNormalization'**

src/Models/Options/ABCNetOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+        if (InputChannels <= 0) throw new ArgumentOutOfRangeException(nameof(InputChannels), InputChannels, "InputChannels must be positive.");
+        if (FeatureChannels <= 0) throw new ArgumentOutOfRangeException(nameof(FeatureChannels), FeatureChannels, "FeatureChannels must be positive.");
+        if (FeatureStride <= 0) throw new ArgumentOutOfRangeException(nameof(FeatureStride), FeatureStride, "FeatureStride must be positive.");
+        if (BezierSampleHeight <= 0) throw new ArgumentOutOfRangeException(nameof(BezierSampleHeight), BezierSampleHeight, "BezierSampleHeight must be positive.");
+        if (BezierSampleWidth <= 0) throw new ArgumentOutOfRangeException(nameof(BezierSampleWidth), BezierSampleWidth, "BezierSampleWidth must be positive.");
+        if (RecognitionHiddenSize <= 0) throw new ArgumentOutOfRangeException(nameof(RecognitionHiddenSize), RecognitionHiddenSize, "RecognitionHiddenSize must be positive.");
+        if (MaxInstances <= 0) throw new ArgumentOutOfRangeException(nameof(MaxInstances), MaxInstances, "MaxInstances must be positive.");
```
ooples, line 123: _🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_

**Validate() checks every numeric option except ConfidenceThreshold, the one whose out-of-range value fails silently.**

Lines 95-103 range-check all nine integer options, but `ConfidenceThreshold` (line 78) is never validated. It is a probability consumed at `ABCNet.cs:541` as `if (score < _options.ConfidenceThreshold) continue;`. A value above 1.0 makes that condition true for every position, so detection returns zero instances on every image - a model that appears to run correctly and silently finds nothing. A negative value is the mirror failure: every position becomes a detection, and `MaxInstances` truncation then returns arbitrary noise. Both are exactly the class of value the method's own summary promises to reject ("values that cannot describe a working model").

<details>
<summary>🔒️ Proposed fix: range-check ConfidenceThreshold alongside the others</summary>

```diff
         if (MaxInstances <= 0) throw new ArgumentOutOfRangeException(nameof(MaxInstances), MaxInstances, "MaxInstances must be positive.");
+        if (ConfidenceThreshold < 0.0 || ConfidenceThreshold > 1.0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(ConfidenceThreshold), ConfidenceThreshold,
+                "ConfidenceThreshold must be in [0, 1]; it is compared against a probability score.");
+        }
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/ABCNetOptions.cs` around lines 103, add a range check for ConfidenceThreshold to Validate(), rejecting values outside [0, 1], matching the ArgumentOutOfRangeException style of the checks immediately above it.
```

</details>

src/Models/Options/ABCNetOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        // CTC cannot emit more labels than it has time steps, so an alphabet is only usable if the
+        // rectified width leaves room for it. This is a real constraint rather than a formality: a
+        // silently over-long alphabet makes long words unrecognizable no matter how well trained.
+        if (NumCharacterClasses < 2)
```
ooples: _📐 Maintainability & Code Quality_ | _🟡 Minor_ | _⚡ Quick win_

**The comment above this check describes a CTC time-step constraint that the check does not implement.**

The comment says the alphabet "is only usable if the rectified width leaves room for it" and insists it is "a real constraint rather than a formality". The check beneath it only rejects `NumCharacterClasses < 2`; nothing relates `NumCharacterClasses` to `BezierSampleWidth`. The reasoning is also inverted - CTC time steps bound the decoded label SEQUENCE length, not the alphabet size, so no relation between the two is required in the first place. As written the comment tells the next reader a guard exists that does not, which is worse than no comment.

<details>
<summary>♻️ Proposed refactor: describe the check that is actually performed</summary>

```diff
-        // CTC cannot emit more labels than it has time steps, so an alphabet is only usable if the
-        // rectified width leaves room for it. This is a real constraint rather than a formality: a
-        // silently over-long alphabet makes long words unrecognizable no matter how well trained.
+        // CTC reserves one label for the blank, so an alphabet needs at least one real symbol
+        // alongside it. (Sequence length is bounded by BezierSampleWidth time steps, but that
+        // constrains how many characters an instance can DECODE, not how large the alphabet is.)
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/ABCNetOptions.cs` around lines 105 - 108, rewrite the comment above the NumCharacterClasses check so it describes the blank-label constraint the code actually enforces, and drop the claim about rectified width bounding alphabet size.
```

</details>

src/Models/Options/BayesianNetworkSynthOptions.cs
```diff
@@ -64,5 +64,55 @@ public class BayesianNetworkSynthOptions<T> : RiskModelOptions<T>
...
+    /// <see cref="StructureBudgetFraction"/>.
+    /// </remarks>
+    public double PrivacyBudget { get; set; } = 1.0;
+
+    /// <summary>
+    /// Gets or sets the fraction of the total privacy budget spent on learning the network structure,
+    /// with the remainder spent on noising the conditional distributions.
+    /// </summary>
+    /// <value>Defaults to 0.5, the paper's even epsilon/2 split between its two phases.</value>
+    public double StructureBudgetFraction { get; set; } = 0.5;
```
ooples: _🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_

**Bound PrivacyBudget and StructureBudgetFraction so a degenerate split cannot silently disable the privacy noise.**

Both new properties are unbounded doubles the generator consumes directly as epsilon shares. In `BayesianNetworkSynthGenerator.EstimateCPTs` the phase-2 budget is `double eps2 = _options.PrivacyBudget * (1.0 - _options.StructureBudgetFraction);` and the scale is `n > 0 && eps2 > 0 ? 2.0 * marginalCount / (n * eps2) : 0.0`. So `StructureBudgetFraction >= 1.0` (or `PrivacyBudget <= 0`) makes `eps2 <= 0`, noiseScale becomes 0.0, and `SampleLaplace(0)` returns 0 for every cell — the DP branch then writes raw empirical marginals into the CPTs, with LaplaceSmoothing also skipped (it only runs in the else branch). The result is an exact, unnoised release of the conditional distributions while `EnableDifferentialPrivacy` is still true, with no exception and no warning.

The structure phase degrades the same way: `stepBudget = PrivacyBudget * StructureBudgetFraction / (d-1)`, and `SampleExponentialMechanism` falls back to a uniform draw when epsilon <= 0, so a zero fraction costs utility silently rather than erroring.

This contradicts the contract stated at lines 81-83 of this file — that running without the privacy mechanisms "has to be an explicit opt-out rather than the default" — since a single out-of-range fraction is an implicit opt-out. A non-positive epsilon should mean maximum noise, never zero noise.

<details>
<summary>🔒️ Proposed fix: reject non-positive epsilon and an out-of-range budget split</summary>

```diff
-    public double PrivacyBudget { get; set; } = 1.0;
+    public double PrivacyBudget
+    {
+        get => _privacyBudget;
+        set
+        {
+            if (double.IsNaN(value) || value <= 0.0)
+                throw new ArgumentOutOfRangeException(nameof(value), value,
+                    "PrivacyBudget (epsilon) must be greater than zero; a non-positive budget silently removes the noise instead of maximizing it.");
+            _privacyBudget = value;
+        }
+    }
+    private double _privacyBudget = 1.0;

-    public double StructureBudgetFraction { get; set; } = 0.5;
+    public double StructureBudgetFraction
+    {
+        get => _structureBudgetFraction;
+        set
+        {
+            if (double.IsNaN(value) || value <= 0.0 || value >= 1.0)
+                throw new ArgumentOutOfRangeException(nameof(value), value,
+                    "StructureBudgetFraction must be strictly between 0 and 1 so both PrivBayes phases receive a positive share of epsilon.");
+            _structureBudgetFraction = value;
+        }
+    }
+    private double _structureBudgetFraction = 0.5;
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/BayesianNetworkSynthOptions.cs` around lines 115, in src/Models/Options/BayesianNetworkSynthOptions.cs, convert PrivacyBudget and StructureBudgetFraction to properties with backing fields and validating setters: PrivacyBudget must be > 0 and not NaN, StructureBudgetFraction must be strictly between 0 and 1 and not NaN, both throwing ArgumentOutOfRangeException explaining that a degenerate value zeroes the phase-2 laplace noise in BayesianNetworkSynthGenerator.EstimateCPTs while EnableDifferentialPrivacy is still true. keep the existing xml doc comments and the 1.0 and 0.5 defaults.
```

</details>

src/Models/Options/ChronosFinanceOptions.cs
```diff
@@ -125,6 +126,23 @@ public ChronosFinanceOptions(ChronosFinanceOptions<T> other)
...
```
ooples, line 144: _🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_

**Validate QuantizationBound in Chronos.ValidateOptions alongside its siblings.**

QuantizationBound is the only numeric option on this class that nothing bounds. `Chronos.ValidateOptions` (src/Finance/Forecasting/Foundation/Chronos.cs:462-483) checks ContextLength, ForecastHorizon, NumTokens, HiddenDimension, NumLayers, NumHeads, IntermediateSize, NumSamples, DropoutRate and Temperature — every value this file declares except the one added here.

It is consumed at Chronos.cs:1023-1025 as the quantizer's divisor: `_lastTokenRange = 2.0 * bound * scale`. With `QuantizationBound = 0` the range is exactly zero, so line 1072 (`(values[i] - min) / range`) divides by zero for every element. Nothing throws: the resulting NaN/Infinity is cast to int and then clamped by `Math.Min(..., _numTokens - 1)` / `Math.Max(0, ...)` at 1075-1076, so every timestep collapses to bin 0, and Detokenize (line 1123, `scaledValue * rangeVal + minVal` with both zero) returns an all-zero forecast with no error.

A negative bound is equally unguarded and silently inverts the bin ordering (min becomes +|b|s, range becomes -2|b|s), so the largest values map to the lowest token ids — destroying the ordinal token structure the model is trained on. Both are silent wrong-output paths reachable purely from a user-set option, which is why the sibling values are checked.

<details>
<summary>🔒️ Proposed fix: bound QuantizationBound in ValidateOptions</summary>

```diff
         if (options.Temperature <= 0)
             errors.Add("Temperature must be positive.");
+        if (options.QuantizationBound <= 0 ||
+            double.IsNaN(options.QuantizationBound) ||
+            double.IsInfinity(options.QuantizationBound))
+            errors.Add("QuantizationBound must be a finite positive value.");
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/ChronosFinanceOptions.cs` around lines 144, in src/Finance/Forecasting/Foundation/Chronos.cs, add a QuantizationBound check to ValidateOptions next to the existing Temperature check: reject values that are non-positive, NaN, or infinite with the message "QuantizationBound must be a finite positive value.", since the value becomes the quantizer divisor at CaptureTokenScale and a zero or negative bound silently collapses or inverts every token bin.
```

</details>

src/Models/Options/DiffusionModelOptions.cs
```diff
@@ -105,6 +107,32 @@ public DiffusionModelOptions(DiffusionModelOptions<T> other)
...
+    /// </summary>
+    /// <value>
+    /// The latent channel count, or <c>null</c> to use the model's own architectural default.
+    /// </value>
+    /// <remarks>
+    /// <para>
+    /// Nullable on purpose, following the options pattern: <c>null</c> means "use the value this
+    /// architecture was designed around" rather than forcing every caller to know it. Latent
+    /// diffusion models are not free to pick this arbitrarily — it must match the VAE and the
+    /// predictor's input/output channels — so a model that cannot honour an override should say so
```
ooples, line 121: _🩺 Stability & Availability_ | _🟡 Minor_ | _⚡ Quick win_

**Give UniVSTModel the same LatentChannels guards TLoRAModel already has, or the documented contract is unmet.**

The remark on the new property (lines 118-122) states that an override "must match the VAE and the predictor's input/output channels — so a model that cannot honour an override should say so rather than silently ignore it". Of the two models this PR wires to the option, only one does.

TLoRAModel.cs:118-135 rejects a non-positive value with `ArgumentOutOfRangeException` and throws `ArgumentException` when an injected VAE's LatentChannels disagrees with the override. UniVSTModel.cs:150-151 does neither: `int requested = options?.LatentChannels ?? DEFAULT_LATENT_CHANNELS; _latentChannels = vae?.LatentChannels ?? requested;`.

Concretely, `new UniVSTModel<float>(options: new DiffusionModelOptions<float> { LatentChannels = 0 })` with no VAE sets `_latentChannels = 0`, which flows into InitializeLayers (168-174) as UNetNoisePredictor input/output channels = 0 and StandardVAE latentChannels = 0. StandardVAE's constructor (src/Diffusion/VAE/StandardVAE.cs:259-272) assigns it straight to a field with no positivity check, so the value only surfaces as a zero-width conv construction failure far from the option that caused it. Second case: passing a 4-channel VAE together with `LatentChannels = 8` silently discards the override on line 151 — exactly the "silently ignore" the doc says must not happen, while the identical call on TLoRAModel throws.

<details>
<summary>🔒️ Proposed fix: validate LatentChannels in UniVSTModel</summary>

```diff
         int requested = options?.LatentChannels ?? DEFAULT_LATENT_CHANNELS;
+        if (requested <= 0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(options), requested,
+                "DiffusionModelOptions.LatentChannels must be positive when specified.");
+        }
+
+        if (vae is not null && options?.LatentChannels is int overridden && vae.LatentChannels != overridden)
+        {
+            throw new ArgumentException(
+                $"DiffusionModelOptions.LatentChannels was set to {overridden}, but the supplied VAE " +
+                $"produces {vae.LatentChannels} latent channels. Either omit the override or supply a " +
+                "VAE built for the same width.", nameof(options));
+        }
+
         _latentChannels = vae?.LatentChannels ?? requested;
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/DiffusionModelOptions.cs` around lines 121, in src/Diffusion/StyleTransfer/UniVSTModel.cs, mirror the LatentChannels validation TLoRAModel.cs performs: after computing requested at line 150, throw ArgumentOutOfRangeException when it is <= 0, and throw ArgumentException when a vae is supplied and options.LatentChannels disagrees with vae.LatentChannels, instead of silently letting the vae win.
```

</details>

src/Models/Options/ExtraTreesClassifierOptions.cs
```diff
@@ -62,12 +62,27 @@ public class ExtraTreesClassifierOptions<T> : ClassifierOptions<T>
...
+    /// A positive count, or <c>null</c> (the default) to use the <see cref="MaxFeatures"/> rule.
     /// </value>
-    public string MaxFeatures { get; set; } = "sqrt";
+    /// <remarks>
+    /// A count is a number, so it gets its own numeric property rather than being smuggled through
+    /// the rule. The previous <c>string MaxFeatures</c> had to accept both, and its parse fell back
+    /// to sqrt on anything unrecognized — so a typo silently trained a different model.
+    /// Clamped to the available feature count at use.
+    /// </remarks>
+    public int? MaxFeatureCount { get; set; }
```
ooples, line 85: _🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_

**Copy MaxFeatureCount in ExtraTreesClassifier.Clone and CreateNewInstance so cloned models keep the explicit feature count.**

`MaxFeatureCount` (line 85) is new in this diff and is documented at lines 74-77 as overriding `MaxFeatures`. `CalculateMaxFeatures` honours that override (ExtraTreesClassifier.cs:195-203 returns `Math.Min(explicitCount, NumFeatures)` before the enum switch), but neither option-copy site carries the new property: `CreateNewInstance` (269-280) and `Clone` (286-297) both copy `MaxFeatures = Options.MaxFeatures` and stop. There is no `MaxFeatureCount = Options.MaxFeatureCount` anywhere in the repo, so nothing compensates.

Concrete failure: configure `new ExtraTreesClassifierOptions<double> { MaxFeatureCount = 3 }` on 100 features and train — CalculateMaxFeatures returns 3, correct. Now take the model through anything that goes via CreateNewInstance (cross-validation refits, hyperparameter search, ensemble wrappers that clone a prototype before retraining): the fresh instance has `MaxFeatureCount == null`, falls through to `MaxFeatures`, still at its `Sqrt` default, and retrains each tree on `ceil(sqrt(100)) == 10` features instead of 3. No exception, no warning — the refit model is simply a different model from the one configured, and CV scores describe a configuration the user never asked for. `Clone()` has the same hole and additionally clones already-trained estimators, so the object reports a configuration inconsistent with the trees it carries.

This is the silent-substitution failure the remarks at lines 80-83 say the redesign exists to eliminate. RandomForestClassifier.cs:384/403 has the identical omission for its own MaxFeatureCount.

<details>
<summary>🔒️ Proposed fix: propagate MaxFeatureCount on clone and refit</summary>

```diff
             MaxFeatures = Options.MaxFeatures,
+            MaxFeatureCount = Options.MaxFeatureCount,
             Criterion = Options.Criterion,
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/ExtraTreesClassifierOptions.cs` around lines 85, in src/Classification/Ensemble/ExtraTreesClassifier.cs, add `MaxFeatureCount = Options.MaxFeatureCount,` to the options object literals in both CreateNewInstance (around line 275) and Clone (around line 292), immediately after the existing MaxFeatures line, so an explicit per-split feature count survives a clone or refit instead of silently reverting to the Sqrt rule. apply the identical two-line fix to RandomForestClassifier.cs at lines 384 and 403, which has the same omission.
```

</details>

ooples (commented):
**Actionable comments posted: 6**

> [!WARNING]
> **Partial review: 6 of 96 files reviewed — 89 not reviewed.**
> Findings below cover only the reviewed files. The remainder have not been looked at,
> so silence about them is not a clean bill of health.

<details>
<summary>🧹 Nitpick comments (1)</summary><blockquote>

<details>
<summary>src/Models/Options/ABCNetOptions.cs (1)</summary><blockquote>

`20`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_

**The type parameter T is declared and documented but never used in the class body.**

`ABCNetOptions<T>` has no member that mentions `T`; the base `NeuralNetworkOptions` is non-generic and every property is `int` or `double`. Callers must still write `new ABCNetOptions<double>()`. If the surrounding Options types carry an unused `T` by convention this is fine as-is - worth confirming before changing, since dropping it is a source-breaking change for callers.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/Models/Options/ABCNetOptions.cs` around lines 20, confirm whether sibling Options types in src/Models/Options carry an unused type parameter by convention; if not, consider dropping T from ABCNetOptions, noting it is a source-breaking change.
```

</details>



</blockquote></details>

</blockquote></details>

<details>
<summary>🤖 Prompt for all review comments with AI agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

Inline comments:
In @src/Models/Options/ABCNetOptions.cs:
- Around line 103: add a range check for ConfidenceThreshold to Validate(), rejecting values outside [0, 1], matching the ArgumentOutOfRangeException style of the checks immediately above it.
- Around line 105-108: rewrite the comment above the NumCharacterClasses check so it describes the blank-label constraint the code actually enforces, and drop the claim about rectified width bounding alphabet size.
- Around line 20: confirm whether sibling Options types in src/Models/Options carry an unused type parameter by convention; if not, consider dropping T from ABCNetOptions, noting it is a source-breaking change.

In @src/Models/Options/BayesianNetworkSynthOptions.cs:
- Around line 115: in src/Models/Options/BayesianNetworkSynthOptions.cs, convert PrivacyBudget and StructureBudgetFraction to properties with backing fields and validating setters: PrivacyBudget must be > 0 and not NaN, StructureBudgetFraction must be strictly between 0 and 1 and not NaN, both throwing ArgumentOutOfRangeException explaining that a degenerate value zeroes the phase-2 laplace noise in BayesianNetworkSynthGenerator.EstimateCPTs while EnableDifferentialPrivacy is still true. keep the existing xml doc comments and the 1.0 and 0.5 defaults.

In @src/Models/Options/ChronosFinanceOptions.cs:
- Around line 144: in src/Finance/Forecasting/Foundation/Chronos.cs, add a QuantizationBound check to ValidateOptions next to the existing Temperature check: reject values that are non-positive, NaN, or infinite with the message "QuantizationBound must be a finite positive value.", since the value becomes the quantizer divisor at CaptureTokenScale and a zero or negative bound silently collapses or inverts every token bin.

In @src/Models/Options/DiffusionModelOptions.cs:
- Around line 121: in src/Diffusion/StyleTransfer/UniVSTModel.cs, mirror the LatentChannels validation TLoRAModel.cs performs: after computing requested at line 150, throw ArgumentOutOfRangeException when it is <= 0, and throw ArgumentException when a vae is supplied and options.LatentChannels disagrees with vae.LatentChannels, instead of silently letting the vae win.

In @src/Models/Options/ExtraTreesClassifierOptions.cs:
- Around line 85: in src/Classification/Ensemble/ExtraTreesClassifier.cs, add `MaxFeatureCount = Options.MaxFeatureCount,` to the options object literals in both CreateNewInstance (around line 275) and Clone (around line 292), immediately after the existing MaxFeatures line, so an explicit per-split feature count survives a clone or refit instead of silently reverting to the Sqrt rule. apply the identical two-line fix to RandomForestClassifier.cs at lines 384 and 403, which has the same omission.
```

</details>

<details>
<summary>🪄 Autofix</summary>

Fix all unresolved review comments on this PR:

- [ ] Push a commit to this branch
- [ ] Create a new PR with the fixes

</details>

<details>
<summary>ℹ️ Review info</summary>

<details>
<summary>⚙️ Run configuration</summary>

**Configuration used**: .coderabbit.yaml

**Review profile**: assertive

**Engine**: pr-review plugin (self-hosted)

</details>

<details>
<summary>📥 Commits</summary>

Reviewing files that changed from the base of the PR and between 4185d63e6f52ac802af10e27ad57acdb49473766 and 0842eb921b4ef30217d80d036337b774cdf17d97.

</details>

<details>
<summary>📒 Files reviewed (6 of 96)</summary>

* `src/Models/Options/ABCNetOptions.cs`
* `src/Models/Options/BayesianNetworkSynthOptions.cs`
* `src/Models/Options/CausalDiscoveryOptions.cs`
* `src/Models/Options/ChronosFinanceOptions.cs`
* `src/Models/Options/DiffusionModelOptions.cs`
* `src/Models/Options/ExtraTreesClassifierOptions.cs`

</details>

</details>

ooples:
## Walkthrough

Review of the Options-class slice of #1789. This is a **partial review in progress** — 7 of 96 files have been read so far, and the review body states the shortfall explicitly. Remaining files are queued and will be reviewed in follow-up passes; silence about a file is not yet a clean bill of health for it.

Findings so far cluster on one theme: newly-added options that reach a consumer which does not bound them, while every sibling option on the same class is bounded. In each case the out-of-range value produces a silently wrong result rather than an error — unnoised differential privacy, an all-zero forecast, or a model refit on a different feature count than was configured.

### Changes

**Options validation**

|Layer / File(s)|Summary|
|---|---|
|**Synthetic data / privacy** <br> `src/Models/Options/BayesianNetworkSynthOptions.cs`|New PrivacyBudget and StructureBudgetFraction are unbounded; a degenerate split zeroes the Laplace noise while differential privacy still reports enabled.|
|**Finance forecasting** <br> `src/Models/Options/ChronosFinanceOptions.cs`|New QuantizationBound is the only numeric option ValidateOptions does not check; zero divides by zero in the quantizer and returns an all-zero forecast.|
|**Diffusion** <br> `src/Models/Options/DiffusionModelOptions.cs`|LatentChannels override is validated by TLoRAModel but silently ignored by UniVSTModel, contrary to the property's own documented contract.|
|**Tree ensembles** <br> `src/Models/Options/ExtraTreesClassifierOptions.cs`|New MaxFeatureCount is not copied in Clone or CreateNewInstance, so any refit reverts to the Sqrt default. RandomForestClassifier has the identical omission.|
|**Reviewed, no findings** <br> `src/Models/Options/CausalDiscoveryOptions.cs`, `src/Models/Options/AttentionAllocationOptions.cs`|CausalDiscoveryOptions: both consumers and the copy constructor check out. AttentionAllocationOptions: deleted by this PR, nothing to anchor.|

**Estimated code review effort:** 3 (moderate) | ~40 minutes

src/Models/Options/ChronosFinanceOptions.cs
ooples: Already fixed on this branch, in `4185d63e6` ("refactor(1789): tests + Models (02/18)"). I verified the property is copied rather than trusting the line number, and re-audited the whole constructor: **18/18 properties carried, none missing.**

Same check run across the other three files flagged by AIDN070 on this PR — `DiffusionModelOptions` 12/12 and `RWKVForecastingOptions` 13/13 — and all four flagged properties (`QuantizationBound`, `LatentChannels`, `RevInEpsilon`, `UseReversibleNormalization`) are present in their copy constructors.

Worth noting for the tooling: my first pass at this audit reported five false "not copied" hits on a sibling file, because these constructors pack several assignments onto one line and my matcher was line-anchored. The counts above are from the corrected audit, which also treats a deep copy or an array `Clone()` as a copy rather than an omission.

src/Models/Options/DiffusionModelOptions.cs
ooples: Already fixed on this branch, in `4185d63e6` ("refactor(1789): tests + Models (02/18)"). I verified the property is copied rather than trusting the line number, and re-audited the whole constructor: **18/18 properties carried, none missing.**

Same check run across the other three files flagged by AIDN070 on this PR — `DiffusionModelOptions` 12/12 and `RWKVForecastingOptions` 13/13 — and all four flagged properties (`QuantizationBound`, `LatentChannels`, `RevInEpsilon`, `UseReversibleNormalization`) are present in their copy constructors.

Worth noting for the tooling: my first pass at this audit reported five false "not copied" hits on a sibling file, because these constructors pack several assignments onto one line and my matcher was line-anchored. The counts above are from the corrected audit, which also treats a deep copy or an array `Clone()` as a copy rather than an omission.

src/Models/Options/RWKVForecastingOptions.cs
ooples: Already fixed on this branch, in `4185d63e6` ("refactor(1789): tests + Models (02/18)"). I verified the property is copied rather than trusting the line number, and re-audited the whole constructor: **18/18 properties carried, none missing.**

Same check run across the other three files flagged by AIDN070 on this PR — `DiffusionModelOptions` 12/12 and `RWKVForecastingOptions` 13/13 — and all four flagged properties (`QuantizationBound`, `LatentChannels`, `RevInEpsilon`, `UseReversibleNormalization`) are present in their copy constructors.

Worth noting for the tooling: my first pass at this audit reported five false "not copied" hits on a sibling file, because these constructors pack several assignments onto one line and my matcher was line-anchored. The counts above are from the corrected audit, which also treats a deep copy or an array `Clone()` as a copy rather than an omission.

src/Models/Options/RWKVForecastingOptions.cs
ooples: Already fixed on this branch, in `4185d63e6` ("refactor(1789): tests + Models (02/18)"). I verified the property is copied rather than trusting the line number, and re-audited the whole constructor: **18/18 properties carried, none missing.**

Same check run across the other three files flagged by AIDN070 on this PR — `DiffusionModelOptions` 12/12 and `RWKVForecastingOptions` 13/13 — and all four flagged properties (`QuantizationBound`, `LatentChannels`, `RevInEpsilon`, `UseReversibleNormalization`) are present in their copy constructors.

Worth noting for the tooling: my first pass at this audit reported five false "not copied" hits on a sibling file, because these constructors pack several assignments onto one line and my matcher was line-anchored. The counts above are from the corrected audit, which also treats a deep copy or an array `Clone()` as a copy rather than an omission.

src/Models/Options/ABCNetOptions.cs
ooples: Fixed in `822f3db0c`.

You identified the one that fails silently. `Validate` bounded all nine integers and left the single probability unbounded, and it is compared at `ABCNet.cs:541` as `if (score < _options.ConfidenceThreshold) continue;` — so above 1.0 that test is true at every position and detection returns zero instances on **every** image, which reads as a trained-but-useless model rather than a misconfiguration.

```csharp
if (ConfidenceThreshold < 0.0 || ConfidenceThreshold > 1.0)
{
    throw new ArgumentOutOfRangeException(nameof(ConfidenceThreshold), ConfidenceThreshold,
        "ConfidenceThreshold must be within [0, 1]. Above 1 no detection can ever pass it; "
        + "below 0 every position passes.");
}
```

**This had a consequence worth flagging.** `ABCNet`'s constructor calls `options.Validate()` (`ABCNet.cs:153`), and `AbcNetSpottingTests.RaisingTheThresholdRemovesInstances` deliberately set `ConfidenceThreshold = 1.5` to prove the threshold is actually applied — so the new bound would have thrown there and broken a passing test. The test now uses `1.0`, which preserves its intent exactly: a sigmoid maps to the **open** interval (0, 1) and never reaches 1.0, so the threshold is still unreachable and must still empty the result.

The new `LearningRate` on this class is validated the same way (positive and finite).

src/Models/Options/ABCNetOptions.cs
ooples: Fixed in `822f3db0c`. You were right on both counts — the check does not implement what the comment claimed, and the claim was backwards.

CTC's time-step count bounds the length of a decoded **label sequence**, not the **size** of the alphabet those labels are drawn from: `BezierSampleWidth` time steps can emit at most that many labels whether the alphabet holds 97 characters or 97,000. So there is no width-vs-alphabet relation to enforce here, and the genuine width constraint is about the longest word to be recognized — a property of the data, not of this configuration, so it is not checkable at this point.

The comment now states only what the check enforces (at least one real symbol plus the CTC blank) and records why the previous reasoning was wrong, so it does not get reintroduced.

src/Models/Options/BayesianNetworkSynthOptions.cs
ooples: Fixed in `822f3db0c`.

Your reading of the consequence is what made this worth doing properly: at `StructureBudgetFraction = 1.0` the phase-2 share is exactly zero, the `eps2 > 0` test fails, `noiseScale` stays `0.0`, and the conditional distributions are published with **no privacy noise at all** while the object still reports differential privacy as enabled. That is a privacy failure that looks like a working configuration, which is the worst shape for one to take.

Both are validated in the setters rather than in a `Validate()` method, so the invalid value can never be stored regardless of which consumer reads it — and this slice does not contain the generator's DP code (`eps2` lives on #1789), so a validation hook there would not have covered it here:

```csharp
if (value <= 0.0 || value >= 1.0 || double.IsNaN(value))
{
    throw new ArgumentOutOfRangeException(nameof(StructureBudgetFraction), value,
        "StructureBudgetFraction must be strictly between 0 and 1. At 1 the second phase "
        + "receives no budget and its distributions are published without privacy noise; "
        + "at 0 the structure is learned without any.");
}
```

`PrivacyBudget` is bounded to positive-and-finite on the same reasoning: epsilon divides the noise scale, so a non-positive budget is not a weaker guarantee, it is no guarantee. Defaults (1.0 and 0.5) are unchanged and both remain valid, and no existing assignment in the repo falls outside the new ranges — I checked before tightening.

src/Models/Options/ChronosFinanceOptions.cs
ooples: Fixed in `822f3db0c`. It was the only numeric option on the class that nothing checked while all ten of its siblings were checked, so the inconsistency you describe was exact.

```csharp
// The bound is the half-width of the binning domain, so a non-positive value collapses
// [-bound*s, +bound*s] to a point or inverts it -- every observation then falls in one bin
// and the forecast is constant.
if (options.QuantizationBound <= 0)
    errors.Add("QuantizationBound must be positive.");
```

Added in `Chronos.ValidateOptions` beside the others so it joins the same aggregated `errors` list and reports through the one `ArgumentException`, rather than throwing early and hiding any other invalid option in the same configuration.

src/Models/Options/DiffusionModelOptions.cs
ooples: Fixed on **#1972 in `dceb8961f`** — the finding is correct, but `UniVSTModel` lives in `src/Diffusion/StyleTransfer/`, which this slice does not contain; only the option it violates is here.

Your diagnosis held up exactly. The constructor read

```csharp
int requested = options?.LatentChannels ?? DEFAULT_LATENT_CHANNELS;
_latentChannels = vae?.LatentChannels ?? requested;
```

so a caller who set the option **and** supplied a VAE had their value discarded without a word, and a non-positive override was accepted outright. Both guards are now the same two `TLoRAModel` already carries — reject a non-positive value, and reject an override that disagrees with the supplied VAE's width rather than letting it surface as a shape mismatch deep inside the denoising loop. The documented contract on this property is met by both models now, so the remark stands as written.

src/Models/Options/ExtraTreesClassifierOptions.cs
ooples: Fixed in `822f3db0c`.

Confirmed against `CalculateMaxFeatures` (`ExtraTreesClassifier.cs:195-203`), which returns `Math.Min(explicitCount, NumFeatures)` before reaching the enum switch — so an omitted `MaxFeatureCount` does not merely lose a value, it silently demotes an explicit feature count back to the `MaxFeatures` enum on every copy. Both sites carried the other nine options, which is what made the gap easy to miss.

`MaxFeatureCount = Options.MaxFeatureCount` is now in **both** `CreateNewInstance` and `Clone`; I checked there were exactly two option-copy sites rather than fixing only the one on the flagged line.

PR #1969, opened 2026-08-07
refactor(1789): NeuralNetworks (04/18)

ooples:
Slice 4 of 18 splitting #1789 into review-sized PRs.

**Scope:** NeuralNetworks

**Files:** 96

## Why an integration branch

## Integrity

Related: #1789


src/NeuralNetworks/Layers/LayerBase.cs
```diff
@@ -1591,7 +2015,50 @@ public virtual Tensor<T> ForwardGpu(IReadOnlyDictionary<string, Tensor<T>> input
...
+    /// </summary>
+    /// <param name="input">Input tensor.</param>
+    /// <returns>The transformed tensor.</returns>
+    /// <remarks>
+    /// Virtual rather than abstract purely so the migration off <see cref="Forward"/> can proceed one
+    /// layer at a time without a broken build in between. The default throws, because a layer that
+    /// implements neither has no computation at all - that is a bug, not a default worth inventing.
+    /// </remarks>
+    protected virtual Tensor<T> ForwardTraced(Tensor<T> input)
+        => throw new NotImplementedException(
```
ooples: _🐛 Correctness_ | _🔴 Critical_

**BLOCKING: 'ForwardTraced' is a stub**

Throwing NotImplementedException ships a method that fails at runtime. Implement it, or throw NotSupportedException when the operation genuinely does not apply to this type.

Flagged by `AIDN074` (Correctness). This is a **regression**: the rule does not fire on this file on `master`, so this PR introduces it.

ooples (commented):

src/NeuralNetworks/Layers/LayerBase.cs
ooples: Fixed, taking your second option — `NotSupportedException`:

```csharp
protected virtual Tensor<T> ForwardTraced(Tensor<T> input)
    => throw new NotSupportedException(
        $"{GetType().Name} implements neither Forward nor ForwardTraced, so it has no forward "
        + "computation. Override ForwardTraced in that layer with its computation; LayerBase "
        + "cannot supply one.");
```

That is the right one of the two, and the reasoning is now in an `<exception>` block rather than left implicit: `NotImplementedException` reads as *"this library has not finished building this yet"* — a promise that some future version supplies a base-class forward pass. There is nothing to supply. A layer with no computation of its own cannot be given one generically, so the condition is permanent and belongs to the **derived** layer, not to `LayerBase`.

I also rewrote the message to say what to do and where, instead of restating the rule.

Two checks before changing the type:

- Nothing under `src/NeuralNetworks` catches `NotImplementedException`, so no handler depended on the old type.
- Build baseline is byte-identical with and without the change — 76 `CS0103`, 20 `CS0104`, 8 `CS0111`, 2 `CS0234`, 468 `CS0246`, all pre-existing cross-slice artifacts of the split.

The `virtual`-with-a-throwing-default shape stays, since the existing remarks explain it: it lets the migration off `Forward` proceed one layer at a time without a broken build in between.

PR #1972, opened 2026-08-07
refactor(1789): Diffusion + LoRA + VisionLanguage (07/18)

ooples:
Slice 7 of 18 splitting #1789 into review-sized PRs.

**Scope:** Diffusion + LoRA + VisionLanguage

**Files:** 96

## Why an integration branch

## Integrity

Related: #1789


src/Diffusion/StyleTransfer/TLoRAModel.cs
```diff
@@ -13,97 +13,320 @@
...
+    /// The horizon handed to each adapter is the SCHEDULER's TrainTimesteps, not the options' value.
+    /// The scheduler is what the denoising loop and the noise-prediction path actually consult, and
+    /// the two can differ; taking the schedule from anywhere else would mask at timesteps the model
+    /// never visits.
+    /// </para>
+    /// </remarks>
+    private IReadOnlyList<TLoRAAttentionAdapter<T>> InjectAdapters(int? seed)
+    {
+        var adapters = new List<TLoRAAttentionAdapter<T>>();
+        var random = seed.HasValue ? new Random(seed.Value) : new Random();
```
ooples: _🔐 Security_ | _🟠 Major_  | _⚡ Quick win_

**BLOCKING: 'new Random(...)' is not cryptographically secure and is seeded from the clock; use RandomHelper.CreateSeededRandom or RandomHelper.CreateSecureRandom**

System.Random is not cryptographically secure and its default seed comes from the clock, so instances constructed close together can share a seed and emit identical streams. Use RandomHelper.CreateSeededRandom(seed) when reproducibility is required, RandomHelper.CreateSecureRandom() otherwise.

Flagged by `AIDN072` (Security). This is a **regression**: the rule does not fire on this file on `master`, so this PR introduces it.

ooples: _🔐 Security_ | _🟠 Major_  | _⚡ Quick win_

**BLOCKING: 'new Random(...)' is not cryptographically secure and is seeded from the clock; use RandomHelper.CreateSeededRandom or RandomHelper.CreateSecureRandom**

System.Random is not cryptographically secure and its default seed comes from the clock, so instances constructed close together can share a seed and emit identical streams. Use RandomHelper.CreateSeededRandom(seed) when reproducibility is required, RandomHelper.CreateSecureRandom() otherwise.

Flagged by `AIDN072` (Security). This is a **regression**: the rule does not fire on this file on `master`, so this PR introduces it.

ooples (commented):
**Actionable comments posted: 2**

src/Diffusion/StyleTransfer/TLoRAModel.cs
ooples: Already fixed on this branch — the thread was never resolved.

`src/Diffusion/StyleTransfer/TLoRAModel.cs` now reads:

```csharp
var random = seed.HasValue
    ? RandomHelper.CreateSeededRandom(seed.Value)
    : RandomHelper.CreateSecureRandom();
```

which is exactly the shape your finding asks for — the seeded branch when reproducibility is required, the secure one otherwise. `git log` dates it to this branch's own commit `a16fe75cb` ("fix(security): use randomhelper instead of new random"), landed after the review.

Verified across the slice's whole scope rather than the flagged line: **zero occurrences of `new Random(`** in `src/Diffusion` (422 files), `src/LoRA`, or `src/VisionLanguage`.

Both threads on this file report the same construct, so this reply applies to both.

ooples: Already fixed on this branch — the thread was never resolved.

`src/Diffusion/StyleTransfer/TLoRAModel.cs` now reads:

```csharp
var random = seed.HasValue
    ? RandomHelper.CreateSeededRandom(seed.Value)
    : RandomHelper.CreateSecureRandom();
```

which is exactly the shape your finding asks for — the seeded branch when reproducibility is required, the secure one otherwise. `git log` dates it to this branch's own commit `a16fe75cb` ("fix(security): use randomhelper instead of new random"), landed after the review.

Verified across the slice's whole scope rather than the flagged line: **zero occurrences of `new Random(`** in `src/Diffusion` (422 files), `src/LoRA`, or `src/VisionLanguage`.

Both threads on this file report the same construct, so this reply applies to both.

PR #1974, opened 2026-08-07
refactor(1789): VisionLanguage + Document + NER + ComputerVision (09/18)

ooples:
Slice 9 of 18 splitting #1789 into review-sized PRs.

**Scope:** VisionLanguage + Document + NER + ComputerVision

**Files:** 96

## Why an integration branch

## Integrity

Related: #1789

### DIFF
--- a/src/ComputerVision/OCR/EndToEnd/ABCNet.cs
+++ b/src/ComputerVision/OCR/EndToEnd/ABCNet.cs
@@ -141,7 +141,18 @@ public class ABCNet<T> : NeuralNetworkBase<T>, ICompositeLoss<T>
     {
         _options = options;
         _lossFunction = lossFunction ?? new MeanSquaredErrorLoss<T>();
-        _optimizer = optimizer ?? new AdamOptimizer<T, Tensor<T>, Tensor<T>>(this);
+        // THE RATE IS NOW THE OPTIONS' RATHER THAN AdamOptimizer'S GENERIC DEFAULT. Constructed bare,
+        // this trained at InitialLearningRate = 0.001 with no way for a caller to change it short of
+        // building the whole optimizer. ABCNetOptions.LearningRate now supplies it, defaulting to the
+        // paper's 0.01 -- see the remark on that property: 0.01 is the paper's SGD+momentum rate and is
+        // well above the usual Adam scale, so reproducing the paper means injecting an SGD optimizer
+        // here, and staying on this default Adam optimizer means lowering the rate to around 1e-4.
+        _optimizer = optimizer ?? new AdamOptimizer<T, Tensor<T>, Tensor<T>>(
+            this,
+            new AdamOptimizerOptions<T, Tensor<T>, Tensor<T>>
+            {
+                InitialLearningRate = _options.LearningRate,
+            });
 
         InitializeLayers();
     }
--- a/src/ComputerVision/OCR/EndToEnd/ABCNet.cs
+++ b/src/ComputerVision/OCR/EndToEnd/ABCNet.cs
@@ -1,0 +1,844 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using AiDotNet.Attributes;
+using AiDotNet.Enums;
+using AiDotNet.Helpers;
+using AiDotNet.Interfaces;
+using AiDotNet.LinearAlgebra;
+using AiDotNet.LossFunctions;
+using AiDotNet.Models;
+using AiDotNet.Models.Options;
+using AiDotNet.NeuralNetworks;
+using AiDotNet.NeuralNetworks.Graph;
+using AiDotNet.NeuralNetworks.Layers;
+using AiDotNet.Tensors.Engines;
+using AiDotNet.Tensors.LinearAlgebra;
+
+namespace AiDotNet.ComputerVision.OCR.EndToEnd;
+
+/// <summary>
+/// ABCNet: single-shot scene text spotting that regresses a cubic Bezier boundary per text instance and
+/// rectifies each instance with BezierAlign before recognizing it.
+/// </summary>
+/// <typeparam name="T">The numeric type used for calculations.</typeparam>
+/// <remarks>
+/// <para>
+/// Liu, Chen, Shen, He, Jin and Wang, "ABCNet: Real-time Scene Text Spotting with Adaptive Bezier-Curve
+/// Network" (CVPR 2020 oral, pp. 9809-9818, arXiv:2002.10200).
+/// </para>
+/// <para>
+/// The two contributions live in <see cref="CubicBezierCurve{T}"/> and <see cref="BezierAlign"/>, each
+/// independently testable; this class is what composes them into a spotter. What makes it ABCNet rather
+/// than a generic two-stage pipeline is that the SAME feature map feeds both branches and the curve
+/// parameters are what connect them: the detection head regresses eight control points, and BezierAlign
+/// reads the recognition branch's input through those very control points. A pipeline that detects a box,
+/// crops the image, and hands the crop to a separate recognizer — which is what
+/// <see cref="SceneTextReader{T}"/> does, honestly and usefully — shares no gradient between the two and
+/// cannot represent a curved instance at all.
+/// </para>
+/// <para>
+/// SINGLE SHOT, and that is the source of the paper's real-time claim: text/not-text scores and the 16
+/// curve coordinates come from two 1x1 convolutions over one shared feature map, with no proposal stage
+/// and no per-anchor refinement.
+/// </para>
+/// <para>
+/// The recognition branch is deliberately lightweight — convolutions plus a per-column classifier over
+/// the rectified strip, trained with CTC. The paper's argument is that BezierAlign has already done the
+/// hard part: once the instance is straightened, an attention-based recognizer earns very little over a
+/// cheap one, and its cost would remove the speed advantage.
+/// </para>
+/// <para><b>For Beginners:</b> Text in real photos is often curved — around a logo, along a sign. ABCNet
+/// describes each piece of text with a smooth curve along its top and bottom edges, uses those curves to
+/// straighten the text out, and then reads the straightened version. It finds and reads text in one pass
+/// rather than two separate stages.</para>
+/// </remarks>
+/// <example>
+/// <code>
+/// var model = new ABCNet&lt;double&gt;(new ABCNetOptions&lt;double&gt; { InputHeight = 256, InputWidth = 256 });
+/// foreach (var instance in model.Spot(image))
+///     Console.WriteLine($"score {instance.Score}, {instance.CharacterIndices.Count} characters");
+/// </code>
+/// </example>
+[ModelDomain(ModelDomain.Vision)]
+[ModelCategory(ModelCategory.NeuralNetwork)]
+[ModelTask(ModelTask.Detection)]
+[ModelTask(ModelTask.Classification)]
+[ModelComplexity(ModelComplexity.High)]
+[ModelInput(typeof(Tensor<>), typeof(Tensor<>))]
+[ResearchPaper("ABCNet: Real-time Scene Text Spotting with Adaptive Bezier-Curve Network",
+    "https://arxiv.org/abs/2002.10200",
+    Year = 2020,
+    Authors = "Yuliang Liu, Hao Chen, Chunhua Shen, Tong He, Lianwen Jin, Liangwei Wang")]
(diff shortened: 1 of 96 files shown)


src/Document/VisionLanguage/UDOP.cs
```diff
@@ -252,6 +308,35 @@ protected override void InitializeLayers()
...
+    {
+        if (_lazyShapesWarmed) return;
+        _lazyShapesWarmed = true;
+        if (!_useNativeMode) return;
+
+        var dummy = new Tensor<T>([3, ImageSize, ImageSize]);
+        bool wasTraining = IsTrainingMode;
+        if (wasTraining) SetTrainingMode(false);
+        try { _ = Forward(dummy); }
+        catch { /* best-effort; a real forward failure surfaces on the actual Train/Predict */ }
```
ooples: _🩺 Stability & Availability_ | _🟠 Major_  | _⚡ Quick win_

src/NER/Options/SpanBasedNEROptions.cs
```diff
@@ -118,6 +118,32 @@ public int MaxSequenceLength
...
+    /// <summary>
+    /// Number of layers in each boundary FFNN that produces the span start/end representations.
+    /// </summary>
+    /// <remarks>
+    /// Yu et al.'s released configuration (juntaoy/biaffine-ner, experiments.conf) sets
+    /// <c>ffnn_depth = 2</c> alongside <c>ffnn_size = 150</c>. A single layer halves the depth of
+    /// the transformation that separates a token's start role from its end role, which is the
+    /// distinction the two separate FFNNs exist to draw.
+    /// </remarks>
+    public int FfnnDepth { get; set; } = 2;
```
ooples, line 130: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SpanBasedNEROptions' does not copy property 'FfnnDepth'**

src/NER/Options/SpanBasedNEROptions.cs
```diff
@@ -118,6 +118,32 @@ public int MaxSequenceLength
...
+
+    /// <summary>
+    /// Multiplicative learning-rate decay applied every <see cref="DecayFrequency"/> steps.
+    /// </summary>
+    /// <remarks>
+    /// Yu et al.'s released configuration (juntaoy/biaffine-ner, experiments.conf) sets
+    /// <c>decay_rate = 0.999</c>. The paper itself does not mention a schedule.
+    /// </remarks>
+    public double DecayRate { get; set; } = 0.999;
```
ooples, line 139: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SpanBasedNEROptions' does not copy property 'DecayRate'**

src/NER/Options/SpanBasedNEROptions.cs
```diff
@@ -118,6 +118,32 @@ public int MaxSequenceLength
...
+
+    /// <summary>
+    /// How many optimizer steps between applications of <see cref="DecayRate"/>.
+    /// </summary>
+    /// <remarks>The reference configuration uses <c>decay_frequency = 100</c>.</remarks>
+    public int DecayFrequency { get; set; } = 100;
```
ooples, line 145: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SpanBasedNEROptions' does not copy property 'DecayFrequency'**

ooples (commented):
**Actionable comments posted: 4**

src/ComputerVision/OCR/EndToEnd/ABCNet.cs
```diff
@@ -0,0 +1,844 @@
...
+    public ABCNet(
+        ABCNetOptions<T> options,
+        NeuralNetworkArchitecture<T>? architecture = null,
+        IGradientBasedOptimizer<T, Tensor<T>, Tensor<T>>? optimizer = null,
+        ILossFunction<T>? lossFunction = null)
+        : base(ResolveArchitecture(options, architecture), lossFunction ?? new MeanSquaredErrorLoss<T>())
+    {
+        _options = options;
+        _lossFunction = lossFunction ?? new MeanSquaredErrorLoss<T>();
+        _optimizer = optimizer ?? new AdamOptimizer<T, Tensor<T>, Tensor<T>>(this);
```
ooples: _🐛 Correctness_ | _🟠 Major_ | _⚡ Quick win_

**BLOCKING: the training rate is neither the paper's nor user-overridable.**

The optimizer is constructed bare, so it runs on AdamW's own generic default rather than a rate chosen for this model. I checked the options class and there is no `LearningRate` property at all, so a caller has no way to correct it either — the only escape is injecting a fully-built optimizer.

Both halves of the project's rule are broken here: defaults should be the **paper's**, and every default should stay **user-overridable**.

This is a regression — `master` does not construct this optimizer bare.

Fix: add `LearningRate` to the options class with the paper's value as its default, then read it here. I have not asserted a specific number because I could not verify this model's paper in this pass; take it from the paper's training section.

<details>
<summary>🛡️ Proposed fix: read the rate from options</summary>

```diff
-        _optimizer = optimizer ?? new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(this);
+        _optimizer = optimizer ?? new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(
+            this,
+            new AdamWOptimizerOptions<T, Tensor<T>, Tensor<T>>
+            {
+                InitialLearningRate = _options.LearningRate,
+            });
```

</details>

ooples (commented):

src/NER/Options/SpanBasedNEROptions.cs
ooples: Already addressed on the current head. The copy constructor at `SpanBasedNEROptions.cs:267-288` copies this property (lines 283-285 cover `FfnnDepth`, `DecayRate` and `DecayFrequency`). This comment was raised against an earlier commit; verified against the branch tip before resolving.

src/NER/Options/SpanBasedNEROptions.cs
ooples: Already addressed on the current head. The copy constructor at `SpanBasedNEROptions.cs:267-288` copies this property (lines 283-285 cover `FfnnDepth`, `DecayRate` and `DecayFrequency`). This comment was raised against an earlier commit; verified against the branch tip before resolving.

src/NER/Options/SpanBasedNEROptions.cs
ooples: Already addressed on the current head. The copy constructor at `SpanBasedNEROptions.cs:267-288` copies this property (lines 283-285 cover `FfnnDepth`, `DecayRate` and `DecayFrequency`). This comment was raised against an earlier commit; verified against the branch tip before resolving.

src/Document/VisionLanguage/UDOP.cs
ooples: Fixed in d8096e38. The catch now retains the exception on the instance rather than discarding it.

Not logged, because there is nowhere to log to: AiDotNet is a library and carries no logger — nothing under `src/Document` references `ILogger`, `Debug.WriteLine` or any equivalent. The warm-up is best-effort by design and its failure must not propagate, so keeping the exception where a debugger, a test or a later diagnosis can reach it is the available way to stop the first and cleanest failure being invisible.

src/ComputerVision/OCR/EndToEnd/ABCNet.cs
ooples: Confirmed as a real defect, and **left open** because it cannot be fixed on this branch.

`ABCNetOptions<T>` is referenced throughout `ABCNet.cs` but its declaration exists nowhere in this slice — `git grep 'class ABCNetOptions'` returns nothing here, and it is absent from `split/1789-01-core-harness` and `split/1789-10-computervision-video` as well. So there is no options class on this branch to add `LearningRate` to.

That is the same cross-slice dependency documented in #1990: slices 01 and 03 define types each other consumes, and no merge order of the split produces a compiling tree. This fix needs whichever slice owns `ABCNetOptions`.

I have also deliberately not guessed the paper's learning rate. ABCNet (Liu et al., CVPR 2020) trains under Detectron2 with SGD, while this constructor builds an Adam optimizer, so the paper's SGD rate is not transferable to it. Setting a number I cannot justify would satisfy the analyzer while breaking the rule it enforces.

src/NER/Options/SpanBasedNEROptions.cs
ooples: Already fixed on this branch, in `828a9355a` ("fix(options): copy every property in options copy constructors") — `SpanBasedNEROptions.cs:283-285` carries `FfnnDepth`, `DecayRate` and `DecayFrequency`.

I re-audited the whole copy constructor rather than only the flagged line: all 7 auto-properties and all 11 backing fields are copied, `OnnxOptions` is deep-copied via its own copy constructor, and `LabelNames` is cloned rather than shared — so no property is carried by reference either.

src/NER/Options/SpanBasedNEROptions.cs
ooples: Already fixed on this branch, in `828a9355a` ("fix(options): copy every property in options copy constructors") — `SpanBasedNEROptions.cs:283-285` carries `FfnnDepth`, `DecayRate` and `DecayFrequency`.

I re-audited the whole copy constructor rather than only the flagged line: all 7 auto-properties and all 11 backing fields are copied, `OnnxOptions` is deep-copied via its own copy constructor, and `LabelNames` is cloned rather than shared — so no property is carried by reference either.

src/NER/Options/SpanBasedNEROptions.cs
ooples: Already fixed on this branch, in `828a9355a` ("fix(options): copy every property in options copy constructors") — `SpanBasedNEROptions.cs:283-285` carries `FfnnDepth`, `DecayRate` and `DecayFrequency`.

I re-audited the whole copy constructor rather than only the flagged line: all 7 auto-properties and all 11 backing fields are copied, `OnnxOptions` is deep-copied via its own copy constructor, and `LabelNames` is cloned rather than shared — so no property is carried by reference either.

src/Document/VisionLanguage/UDOP.cs
ooples: Fixed in `499ab7200`.

The catch is best-effort by design — a genuine forward failure surfaces again on the real `Train`/`Predict` — so it is still not rethrown. What was wrong is that `catch { }` discarded the diagnosis too, which meant the later failure carried no hint that the warm-up had already hit the same problem:

```csharp
try { _ = Forward(dummy); }
catch (Exception ex)
{
        $"{nameof(UDOP<T>)}: lazy shape resolution failed on a {ImageSize}x{ImageSize} warm-up "
        + $"pass; shapes stay unresolved until the first real Train/Predict. {ex}");
}
```

src/ComputerVision/OCR/EndToEnd/ABCNet.cs
ooples: Fixed in `499ab7200`.

Both halves were real. `new AdamOptimizer<T, Tensor<T>, Tensor<T>>(this)` passed no options, so training fell through to `AdamOptimizerOptions.InitialLearningRate = 0.001`, and `ABCNetOptions` had no rate at all — so injecting a fully-built optimizer really was the only escape.

The property could not go in this PR: `ABCNetOptions<T>` lives at `src/Models/Options/ABCNetOptions.cs`, which this slice does not contain. It was added on **#1967 in `822f3db0c`**, and this commit consumes it:

```csharp
_optimizer = optimizer ?? new AdamOptimizer<T, Tensor<T>, Tensor<T>>(
    this,
    new AdamOptimizerOptions<T, Tensor<T>, Tensor<T>>
    {
        InitialLearningRate = _options.LearningRate,
    });
```

**One caveat, stated rather than buried.** You asked for the paper's value and the default is now the paper's `0.01` — but ABCNet (arXiv:2002.10200) trains with **SGD + momentum**, and this model defaults to **Adam**, where 0.01 is roughly two orders of magnitude above the usual scale. The published number and the default optimizer are not matched to each other, and no single default can satisfy both. Rather than pick silently, the remark on the property and the comment here both say so outright: inject an SGD-with-momentum optimizer to reproduce the paper, or lower this to ~1e-4 to train with the default Adam. `Validate` also now rejects a non-positive or non-finite rate.

Build: 36 errors before, 36 after, identical — all pre-existing split artifacts.

PR #1975, opened 2026-08-07
refactor(1789): ComputerVision + Video (10/18)

ooples:
Slice 10 of 18 splitting #1789 into review-sized PRs.

**Scope:** ComputerVision + Video

**Files:** 96

## Why an integration branch

## Integrity

Related: #1789


ooples (commented):
**Actionable comments posted: 7**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
```diff
@@ -26,6 +26,61 @@ public SAMOptions(SAMOptions other)
...
+
+    /// <summary>Gets or sets AdamW's numerical-stability epsilon.</summary>
+    public double AdamEpsilon { get; set; } = 1e-8;
+
+    /// <summary>
+    /// Gets or sets the linear-warmup length in optimizer steps. The default, 250, is the Segment
+    /// Anything paper's value (Kirillov et al. 2023, §A Training algorithm): the learning rate above
+    /// is the PEAK reached only after warming up over the first 250 iterations.
+    /// </summary>
+    public int WarmupSteps { get; set; } = 250;
```
ooples, line 66: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SAMOptions' does not copy property 'WarmupSteps'**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
```diff
@@ -26,6 +26,61 @@ public SAMOptions(SAMOptions other)
...
+
+    /// <summary>
+    /// Gets or sets the focal-loss weight in the mask objective. The default, 20, is the paper's
+    /// focal:dice ratio of 20:1 (Kirillov et al. 2023, §3).
+    /// </summary>
+    public double MaskFocalWeight { get; set; } = 20.0;
```
ooples, line 72: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SAMOptions' does not copy property 'MaskFocalWeight'**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
```diff
@@ -26,6 +26,61 @@ public SAMOptions(SAMOptions other)
...
+
+    /// <summary>
+    /// Gets or sets the dice-loss weight in the mask objective. The default, 1, is the paper's
+    /// focal:dice ratio of 20:1.
+    /// </summary>
+    public double MaskDiceWeight { get; set; } = 1.0;
```
ooples, line 78: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SAMOptions' does not copy property 'MaskDiceWeight'**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
```diff
@@ -26,6 +26,61 @@ public SAMOptions(SAMOptions other)
...
+
+    /// <summary>
+    /// Gets or sets focal loss's focusing parameter gamma. The default, 2, is the RetinaNet value
+    /// the Segment Anything paper cites.
+    /// </summary>
+    public double FocalGamma { get; set; } = 2.0;
```
ooples, line 84: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SAMOptions' does not copy property 'FocalGamma'**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
```diff
@@ -26,6 +26,61 @@ public SAMOptions(SAMOptions other)
...
+
+    /// <summary>
+    /// Gets or sets focal loss's class-balance parameter alpha. The default, 0.25, is the RetinaNet
+    /// value the Segment Anything paper cites.
+    /// </summary>
+    public double FocalAlpha { get; set; } = 0.25;
```
ooples, line 90: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'SAMOptions' does not copy property 'FocalAlpha'**

src/ComputerVision/Segmentation/OpenVocabulary/GroundedSAM2.cs
```diff
@@ -253,15 +323,119 @@ protected override void InitializeLayers()
...
+    {
+        if (_lazyShapesWarmed) return;
+        _lazyShapesWarmed = true;
+        if (!_useNativeMode || _customLayers) return;
+
+        var dummy = new Tensor<T>([_channels, _height, _width]);
+        bool wasTraining = IsTrainingMode;
+        if (wasTraining) SetTrainingMode(false);
+        try { _ = Forward(dummy); }
+        catch { /* best-effort; a real forward failure surfaces on the actual Train/Predict */ }
```
ooples: _🩺 Stability & Availability_ | _🟠 Major_  | _⚡ Quick win_

src/Video/Denoising/LiteDVDNet.cs
```diff
@@ -114,15 +168,73 @@ protected override void InitializeLayers()
...
+            if (current.Length > 0)
+            {
+                var damped = new Vector<T>(current.Length);
+                var scale = NumOps.FromDouble(_options.ResidualHeadInitScale);
+                for (int i = 0; i < current.Length; i++)
+                    damped[i] = NumOps.Multiply(current[i], scale);
+                head.SetParameters(damped);
+            }
+        }
+        catch (ArgumentException)
```
ooples: _🩺 Stability & Availability_ | _🟠 Major_  | _⚡ Quick win_

src/ComputerVision/Segmentation/Foundation/MixedQueryTransformer.cs
```diff
@@ -0,0 +1,545 @@
...
+        _height = architecture.InputHeight > 0 ? architecture.InputHeight : 512;
+        _width = architecture.InputWidth > 0 ? architecture.InputWidth : 512;
+        _channels = architecture.InputDepth > 0 ? architecture.InputDepth : 3;
+        _numClasses = numClasses;
+        _numQueries = numQueries;
+        _modelSize = modelSize;
+        _dropRate = dropRate;
+        _useNativeMode = true;
+        _onnxModelPath = null;
+        _optimizer = optimizer ?? new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(this);
```
ooples: _🐛 Correctness_ | _🟠 Major_ | _⚡ Quick win_

**BLOCKING: the training rate is neither the paper's nor user-overridable.**

The optimizer is constructed bare, so it runs on AdamW's own generic default rather than a rate chosen for this model. I checked the options class and there is no `LearningRate` property at all, so a caller has no way to correct it either — the only escape is injecting a fully-built optimizer.

Both halves of the project's rule are broken here: defaults should be the **paper's**, and every default should stay **user-overridable**.

This is a regression — `master` does not construct this optimizer bare.

Fix: add `LearningRate` to the options class with the paper's value as its default, then read it here. I have not asserted a specific number because I could not verify this model's paper in this pass; take it from the paper's training section.

<details>
<summary>🛡️ Proposed fix: read the rate from options</summary>

```diff
-        _optimizer = optimizer ?? new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(this);
+        _optimizer = optimizer ?? new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(
+            this,
+            new AdamWOptimizerOptions<T, Tensor<T>, Tensor<T>>
+            {
+                InitialLearningRate = _options.LearningRate,
+            });
```

</details>

ooples (commented):

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already addressed on the current head. `SAMOptions`' copy constructor copies this property (lines 34-38 cover `WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalAlpha` and `FocalGamma`). Raised against an earlier commit; verified against the branch tip before resolving.

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already addressed on the current head. `SAMOptions`' copy constructor copies this property (lines 34-38 cover `WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalAlpha` and `FocalGamma`). Raised against an earlier commit; verified against the branch tip before resolving.

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already addressed on the current head. `SAMOptions`' copy constructor copies this property (lines 34-38 cover `WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalAlpha` and `FocalGamma`). Raised against an earlier commit; verified against the branch tip before resolving.

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already addressed on the current head. `SAMOptions`' copy constructor copies this property (lines 34-38 cover `WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalAlpha` and `FocalGamma`). Raised against an earlier commit; verified against the branch tip before resolving.

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already addressed on the current head. `SAMOptions`' copy constructor copies this property (lines 34-38 cover `WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalAlpha` and `FocalGamma`). Raised against an earlier commit; verified against the branch tip before resolving.

src/ComputerVision/Segmentation/OpenVocabulary/GroundedSAM2.cs
ooples: Fixed in eae4cfb2. The catch now retains the exception on the instance rather than discarding it.

Not logged, because there is nowhere to log to — AiDotNet is a library and carries no logger. The fallback behaviour is unchanged; only the loss of the exception object is.

src/Video/Denoising/LiteDVDNet.cs
ooples: Fixed in eae4cfb2. The catch now retains the exception on the instance rather than discarding it.

Not logged, because there is nowhere to log to — AiDotNet is a library and carries no logger. The fallback behaviour is unchanged; only the loss of the exception object is.

src/ComputerVision/Segmentation/Foundation/MixedQueryTransformer.cs
ooples: Fixed in eae4cfb2.

`MixedQueryTransformerOptions` had no `LearningRate` and neither does `NeuralNetworkOptions`, so the bare optimizer ran on `AdamWOptimizerOptions`' generic default of **0.001** — ten times the **1e-4** this architecture inherits from the Mask2Former recipe (AdamW, lr 1e-4, weight decay 0.05).

`LearningRate` is added with 1e-4 as its default, **copied in the copy constructor** (omitting it there is the AIDN070 defect this PR was flagged for elsewhere), and read at the construction site. If the intended recipe differs, the value is now a one-line change and is user-overridable either way.

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already fixed on this branch — `SAMOptions`'s copy constructor carries all five of the properties flagged across these threads (`WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalGamma`, `FocalAlpha`).

I audited the whole constructor rather than the flagged lines: **10/10 properties copied, none missing and none carried by reference.**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already fixed on this branch — `SAMOptions`'s copy constructor carries all five of the properties flagged across these threads (`WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalGamma`, `FocalAlpha`).

I audited the whole constructor rather than the flagged lines: **10/10 properties copied, none missing and none carried by reference.**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already fixed on this branch — `SAMOptions`'s copy constructor carries all five of the properties flagged across these threads (`WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalGamma`, `FocalAlpha`).

I audited the whole constructor rather than the flagged lines: **10/10 properties copied, none missing and none carried by reference.**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already fixed on this branch — `SAMOptions`'s copy constructor carries all five of the properties flagged across these threads (`WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalGamma`, `FocalAlpha`).

I audited the whole constructor rather than the flagged lines: **10/10 properties copied, none missing and none carried by reference.**

src/ComputerVision/Segmentation/Foundation/SAMOptions.cs
ooples: Already fixed on this branch — `SAMOptions`'s copy constructor carries all five of the properties flagged across these threads (`WarmupSteps`, `MaskFocalWeight`, `MaskDiceWeight`, `FocalGamma`, `FocalAlpha`).

I audited the whole constructor rather than the flagged lines: **10/10 properties copied, none missing and none carried by reference.**

src/ComputerVision/Segmentation/OpenVocabulary/GroundedSAM2.cs
ooples: Fixed in `c7409c5f4`.

The fallback is right — a real forward failure surfaces again on the actual `Train`/`Predict`, so it is still not rethrown — but `catch { }` discarded the diagnosis along with the exception, which left that later failure with no hint the warm-up had already hit the same problem:

```csharp
try { _ = Forward(dummy); }
catch (Exception ex)
{
        $"{nameof(GroundedSAM2<T>)}: lazy shape resolution failed on a "
        + $"{_channels}x{_height}x{_width} warm-up pass; shapes stay unresolved until the "
        + $"first real Train/Predict. {ex}");
}
```

src/Video/Denoising/LiteDVDNet.cs
ooples: Fixed in `c7409c5f4`.

Leaving the head's default initialization in place is the correct fallback, so this still does not rethrow. What the discard cost was the ability to tell two different situations apart: a head that could not be resolved from the declared width and a head that simply had no parameters to damp both ended with `ResidualHeadInitScale` silently not applied, and nothing distinguished them.

```csharp
catch (ArgumentException ex)
{
        $"{nameof(LiteDVDNet<T>)}: could not resolve the residual head from the declared width, "
        + $"so ResidualHeadInitScale was not applied and the head keeps its default "
        + $"initialization. {ex.Message}");
}
```

src/ComputerVision/Segmentation/Foundation/MixedQueryTransformer.cs
ooples: Fixed in `c7409c5f4`. Both halves were as you describe: `new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(this)` passed no options, so training ran at `InitialLearningRate = 1e-3` — ten times the published rate — and `MixedQueryTransformerOptions` had no rate at all, so injecting a fully-built optimizer really was the only escape.

```csharp
/// <value>Defaults to 1e-4, the AdamW rate of the Mask2Former recipe this
/// architecture builds on (arXiv:2404.04469).</value>
public double LearningRate { get; set; } = 1e-4;
```

```csharp
_optimizer = optimizer ?? new AdamWOptimizer<T, Tensor<T>, Tensor<T>>(
    this,
    new AdamWOptimizerOptions<T, Tensor<T>, Tensor<T>>
    {
        InitialLearningRate = _options.LearningRate,
    });
```

Supplying your own optimizer still wins outright; the option is consulted only when the model has to build one.

**One thing worth calling out:** `MixedQueryTransformerOptions` has a copy constructor, so adding a property without adding it there would have recreated the exact AIDN070 bug class the other five findings on this PR are about. `LearningRate = other.LearningRate;` went in at the same time.

Build: 16 errors before, 16 after, identical — all pre-existing split artifacts.

PR #1976, opened 2026-08-07
refactor(1789): Video (11/18)

ooples:
Slice 11 of 18 splitting #1789 into review-sized PRs.

**Scope:** Video

**Files:** 95

## Why an integration branch

## Integrity

Related: #1789


src/Video/Inpainting/E2FGVI.cs
```diff
@@ -849,6 +894,22 @@ protected override Tensor<T> PreprocessFrames(Tensor<T> rawFrames)
...
+        base.ResolveLazyLayerShapes();
+        // The custom flow branch is not traversed by PredictCore, but its lazy
+        // convolution weights are still serialized. Materialize any remaining
+        // layers so clone/round-trip preserves the complete parameter graph.
+        foreach (var layer in Layers)
+        {
+            if (layer is LayerBase<T> lazy && !lazy.IsShapeResolved)
+            {
+                try { lazy.ResolveFromShape([1, 4, Math.Max(8, _height), Math.Max(8, _width)]); }
+                catch (ArgumentException) { }
```
ooples: _🩺 Stability & Availability_ | _🟠 Major_  | _⚡ Quick win_

ooples (commented):

src/Video/Inpainting/E2FGVI.cs
ooples: Fixed:

```csharp
catch (ArgumentException ex)
{
    // NOT SILENT. A layer refusing this probe shape is an EXPECTED outcome ... But swallowing
    // the exception outright made the other case invisible too: a layer that should have
    // resolved and did not just silently drops out of the serialized parameter graph, which is
    // exactly the completeness this override exists to guarantee.
        $"E2FGVI.ResolveLazyLayerShapes: {layer.GetType().Name} rejected the probe " +
        $"shape [1, 4, {Math.Max(8, _height)}, {Math.Max(8, _width)}] and remains " +
        $"lazy; its parameters will not be serialized. {ex.GetType().Name}: {ex.Message}");
}
```

Two notes on the choices, since your suggested remedy names `Logger.LogError/LogWarning`:

**There is no logger at this layer.** I checked before inventing one: no `ILogger` anywhere under `src/Video`, and `NeuralNetworkBase` handles the identical situation with a bare commented catch in three places. What the neighbouring video models *do* use is `System.Diagnostics.Debug.WriteLine` — `SlowFast.cs` in four places — so that is the established channel here rather than a new dependency.

**The message is written to separate the two cases**, which is what makes it worth emitting at all. A layer rejecting the probe is expected and harmless; a layer that *should* have resolved and didn't is a real defect that silently shrinks the serialized parameter graph. Both used to produce identical output — nothing. Naming the layer, the probe shape, and the consequence makes them distinguishable.

Worth flagging: `AIDN076` does not fire on this branch at all (0 occurrences in the build), because the golden-pattern analyzers came to master via #1984 and this slice's base predates them. So I could not verify against the rule itself — the fix addresses the finding on its merits.

Build: 2 `CS0246` before and after, both pre-existing cross-slice artifacts.

PR #1977, opened 2026-08-07
refactor(1789): Video + Audio (12/18)

ooples:
Slice 12 of 18 splitting #1789 into review-sized PRs.

**Scope:** Video + Audio

**Files:** 95

## Why an integration branch

## Integrity

Related: #1789


src/Audio/Effects/RoomImpulseResponse.cs
```diff
@@ -213,10 +532,172 @@ public override void UpdateParameters(Vector<T> parameters)
...
+
+    /// <summary>
+    /// The fixed white-noise realisation shaped into the late field, created on first use.
+    /// </summary>
+    private Tensor<T> GetNoiseSignal(int length)
+    {
+        if (_noiseSignal is not null && _noiseSignal.Shape[^1] == length) return _noiseSignal;
+
+        // Fixed seed: the noise is part of the model, not a per-call random draw. See _noiseSignal.
+        var rng = new Random(20210715);
```
ooples: _🔐 Security_ | _🟠 Major_  | _⚡ Quick win_

**BLOCKING: 'new Random(...)' is not cryptographically secure and is seeded from the clock; use RandomHelper.CreateSeededRandom or RandomHelper.CreateSecureRandom**

System.Random is not cryptographically secure and its default seed comes from the clock, so instances constructed close together can share a seed and emit identical streams. Use RandomHelper.CreateSeededRandom(seed) when reproducibility is required, RandomHelper.CreateSecureRandom() otherwise.

Flagged by `AIDN072` (Security). This is a **regression**: the rule does not fire on this file on `master`, so this PR introduces it.

src/Audio/Enhancement/DCCRN.cs
```diff
@@ -478,6 +514,33 @@ protected override Tensor<T> PostprocessOutput(Tensor<T> modelOutput)
...
+
+        int numBins = _fftSize / 2 + 1;
+        // Enough frames that every strided stage keeps a positive time extent.
+        int frames = Math.Max(4, (int)Math.Pow(2, _numStages));
+        var dummy = new Tensor<T>([1, 2, numBins, frames]);
+
+        bool wasTraining = IsTrainingMode;
+        if (wasTraining) SetTrainingMode(false);
+        try { _ = ForwardNative(dummy); }
+        catch { /* best-effort; a genuine forward failure surfaces on the real Train/Predict */ }
```
ooples: _🩺 Stability & Availability_ | _🟠 Major_  | _⚡ Quick win_

ooples (commented):
**Actionable comments posted: 2**

src/Audio/Effects/RoomImpulseResponse.cs
ooples: Already fixed on this branch — the thread was never resolved.

`src/Audio/Effects/RoomImpulseResponse.cs` now reads:

```csharp
var rng = RandomHelper.CreateSeededRandom(20210715);
```

`git log` dates that to this branch's own commit `8ef595c05` ("fix(security): use randomhelper instead of new random"), which landed after the review comment.

I checked the whole slice rather than just the flagged line: **no `new Random(` remains anywhere under `src/Audio`**. The occurrences still in the tree are in `AiModelBuilder.Configure.cs`, `TimeSeries/TFT`, `ReinforcementLearning` and `NeuralNetworks` — other slices' files, outside this PR's scope.

A seeded rather than secure RNG is the right choice for this one specifically: it generates a room impulse response, where reproducibility is the point, and `CreateSeededRandom` is exactly the branch your finding names for that case.

src/Audio/Enhancement/DCCRN.cs
ooples: Fixed:

```csharp
catch (Exception ex)
{
    // Best-effort by design: this is a warm-up forward whose only job is to materialize lazy
    // shapes ... What the bare `catch { }` cost was the ability to tell those two apart --
    // a warm-up that fails for a REASON (a mis-sized dummy, a stage count the strides cannot
    // support) looked identical to one that succeeded ...
        $"DCCRN warm-up forward failed on a [1, 2, {numBins}, {frames}] probe; lazy shapes " +
        $"may be unresolved. {ex.GetType().Name}: {ex.Message}");
}
```

The exception is still swallowed — that part is deliberate and the existing comment was right about it: a genuine forward failure surfaces again on the real `Train`/`Predict` with a real input. What was wrong was that a warm-up failing *for a reason* and one succeeding produced identical output, so a later failure carried no hint that shape resolution had already gone wrong here. The message includes the probe shape, since a mis-sized dummy is the most likely cause.

`Debug.WriteLine` rather than `Logger.LogError` because there is no logger abstraction at this layer — it is the channel the neighbouring audio and video models already use for this class of diagnostic.

Build: 8 `CS0246` before and after, all pre-existing cross-slice artifacts.

PR #1979, opened 2026-08-07
refactor(1789): SpeechRecognition + TextToSpeech (14/18)

ooples:
Slice 14 of 18 splitting #1789 into review-sized PRs.

**Scope:** SpeechRecognition + TextToSpeech

**Files:** 95

## Why an integration branch

## Integrity

Related: #1789


src/SpeechRecognition/Specialized/CodeSwitchingASROptions.cs
```diff
@@ -26,21 +40,128 @@ public CodeSwitchingASROptions(CodeSwitchingASROptions other)
...
+    /// size.
+    /// </summary>
+    /// <remarks>
+    /// Setting it explicitly overrides the derivation — needed by callers that size the alphabet
+    /// directly, such as bounded test fixtures. An earlier revision made this derived with a no-op
+    /// setter, which SILENTLY discarded those callers' values and left the model building a
+    /// 6001-wide output against a 32-wide fixture. When it is set explicitly the two-inventory
+    /// boundary is still well-defined, via <see cref="MandarinTokenCount"/>.
+    /// </remarks>
+    public int VocabSize
```
ooples, line 157: _🐛 Correctness_ | _🔴 Critical_  | _⚡ Quick win_

**BLOCKING: Copy constructor of 'CodeSwitchingASROptions' does not copy property 'VocabSize'**

ooples (commented):

src/SpeechRecognition/Specialized/CodeSwitchingASROptions.cs
ooples: Checked this one carefully, because the finding is inverted — `VocabSize` **is** copied, and the way it is copied is the actual defect.

The copy constructor had both of these:

```csharp
_explicitVocabSize = other._explicitVocabSize;   // line 43
...
VocabSize = other.VocabSize;                     // line 56
```

`VocabSize` is not an auto-property:

```csharp
public int VocabSize
{
    get => _explicitVocabSize ?? (MandarinCharVocabSize + EnglishBpeVocabSize + 1);
    set => _explicitVocabSize = value;
}
```

So line 56 runs after line 43 and undoes it: the getter **materializes** the derived value, and the setter stores it as an **explicit** one. A clone of a derived-size options object came back pinned — change `MandarinCharVocabSize` on both afterwards and the original recomputes while the clone does not.

Fixed by removing the property assignment. Line 43 already carries both states exactly, which is the whole reason the backing field is copied directly.

This matters more than a normal copy-constructor miss because the class's own remarks record that the explicit-vs-derived distinction has already caused a real defect once — an earlier no-op setter discarded callers' values and left the model building a 6001-wide output against a 32-wide fixture. Collapsing the distinction in the copy constructor is that same failure from the other side.

Two verifications, since the finding cites `AIDN070`:

- `AIDN070` occurrences on this file: **0 before, 0 after**. Removing the assignment does not reintroduce the rule — the analyzer is satisfied by the backing-field copy.
- Build: 4 `CS0246` before and after, both pre-existing cross-slice artifacts.

PR #1981, opened 2026-08-07
refactor(1789): TextToSpeech + Finance (16/18)

ooples:
Slice 16 of 18 splitting #1789 into review-sized PRs.

**Scope:** TextToSpeech + Finance

**Files:** 95

## Why an integration branch

## Integrity

Related: #1789


src/TextToSpeech/Vocoders/PriorGradOptions.cs
```diff
@@ -10,20 +10,48 @@ public class PriorGradOptions : VocoderOptions
...
     public PriorGradOptions()
     {
         SampleRate = 22050;
         MelChannels = 80;
         HopSize = 256;
-        NumDiffusionSteps = 6;
+        NumDiffusionSteps = 50;
+        HiddenDim = 64;
+        LearningRate = 2e-4;
+        WeightDecay = 0.0;
```
ooples, line 34: _🐛 Correctness_ | _🔴 Critical_ | _⚡ Quick win_

**BLOCKING: this silently quadruples PriorGrad's attention heads, from 2 to 8.**

The hardcoded `2` passed to `CreateDefaultDiffusionVocoderLayers` was replaced with `_options.NumHeads`, but `PriorGradOptions()` never assigns `NumHeads`. It therefore inherits `TtsModelOptions.NumHeads = 8`, so every PriorGrad built with default options now gets an 8-head attention stack where it previously got 2.

This is not a refactor, it is a change in the trained architecture: parameter count, memory footprint and any checkpoint written against the old shape all move. Note the same edit handled the sibling constant correctly — `HiddenDim` replaced the hardcoded `64` and the constructor sets `HiddenDim = 64` to preserve it. `NumHeads` was simply missed.

PriorGrad is the only vocoder that reads `_options.NumHeads`, so the fix is contained here.

<details>
<summary>🛡️ Proposed fix: pin the default the code previously used</summary>

```diff
     public PriorGradOptions()
     {
         SampleRate = 22050;
         MelChannels = 80;
         HopSize = 256;
         NumDiffusionSteps = 50;
         HiddenDim = 64;
+        // The layer factory was called with 2 heads before this became configurable;
+        // TtsModelOptions defaults to 8, which would silently change the architecture.
+        NumHeads = 2;
         LearningRate = 2e-4;
         WeightDecay = 0.0;
     }
```

</details>

ooples (commented):

<details>
<summary>🧹 Nitpick comments (2)</summary><blockquote>

**src/TextToSpeech/Vocoders/APNet2.cs**

_🧹 Maintainability_ | _🟡 Minor_ | _⚡ Quick win_

**The base-constructor argument allocates `APNet2Options` three times when `options` is null.**

`(options ?? new APNet2Options())` is evaluated once per loss weight, then `_options = options ?? new APNet2Options()` allocates a fourth. The values agree today because the defaults are constants, so this is not a behavioural bug — but the loss is configured from three instances that are not the one the model keeps, which stops being harmless the moment any default becomes derived or stateful.

<details>
<summary>🛡️ Proposed fix: resolve the options once</summary>

```diff
-        : base(architecture, lossFunction ?? new APNet2GeneratorLoss<T>(
-            (options ?? new APNet2Options()).AmplitudeLossWeight,
-            (options ?? new APNet2Options()).PhaseLossWeight,
-            (options ?? new APNet2Options()).StftLossWeight))
+        : this(architecture, options ?? new APNet2Options(), optimizer, lossFunction, resolved: true)
```

Or add a private static helper that builds the loss from a single resolved instance.

</details>

**src/TextToSpeech/Vocoders/DiffWave.cs**

_🧹 Maintainability_ | _🟡 Minor_

**The optimizer block is detected by remaining-bytes rather than a version tag.**

`reader.BaseStream.Length - reader.BaseStream.Position >= optimizerPayloadBytes` correctly distinguishes old from new payloads **today**: I checked both call sites and `SerializeNetworkSpecificData` is the last thing written on the main path (`NeuralNetworkBase.cs:10551`, immediately followed by `return ms.ToArray()`), and the clone path writes into its own `MemoryStream`. The byte count is also right — 6 doubles plus one int is 52.

The fragility is that this is positional, not self-describing. The moment anything is appended after this section, an old payload will satisfy the length check and 52 bytes of unrelated data will be read as optimizer settings, corrupting everything after it. A leading version int or a bool 'has optimizer block' marker would make it robust and is cheap to add now while the format is still changing.

</blockquote></details>

ooples:
## Walkthrough

The pull request wires configured optimizers and paper-faithful training objectives through the Finance forecasting stack and the TTS vocoders, makes previously hardcoded architecture constants configurable, and extends model serialization to round-trip optimizer settings.

### Changes

**TTS vocoders**

|Layer / File(s)|Summary|
|---|---|
|**Paper objective and optimizer** <br> `src/TextToSpeech/Vocoders/APNet2.cs`, `src/TextToSpeech/Vocoders/APNet2Options.cs`|APNet2 adopts the paper's composite generator loss and AdamW settings (beta1 0.8, beta2 0.99, 2e-4), replacing an inherited squared-error default that penalised phase as an ordinary number.|
|**Optimizer wiring and round-trip** <br> `src/TextToSpeech/Vocoders/DiffWave.cs`, `src/TextToSpeech/Vocoders/PriorGrad.cs`|Training passes the configured optimizer, wraps mode restoration in `try/finally`, and serializes optimizer settings behind a backward-compatible length check.|
|**Configurable architecture** <br> `src/TextToSpeech/Vocoders/PriorGrad.cs`, `src/TextToSpeech/Vocoders/PriorGradOptions.cs`|Hardcoded hidden width and head count become options. See the review comment: the head count changes behaviour.|
|**Vocoder simplification** <br> `src/TextToSpeech/Vocoders/Vocos.cs`, `src/TextToSpeech/VoiceCloning/MetaVoice1B.cs`|Vocos sheds duplicated setup; MetaVoice1B aligns its training path with the shared base.|

**Finance forecasting**

|Layer / File(s)|Summary|
|---|---|
|**State-space models** <br> `src/Finance/Forecasting/StateSpace/TimeMachine.cs`, `src/Finance/Forecasting/StateSpace/Hippo.cs`, `src/Finance/Forecasting/StateSpace/RWKVForecaster.cs`|Optimizers reach the training base through `TrainingOptimizer`, and duplicated setup collapses into the shared base.|
|**Foundation forecasters** <br> `src/Finance/Forecasting/Foundation/SimMTM.cs`, `Chronos.cs`, `YingLong.cs`, `TFC.cs`, `TOTEM.cs`, `MOMENT.cs`, `TimeBridge.cs`, `TOTO.cs`, `TinyTimeMixers.cs`, `Kronos.cs`, `GPT4TS.cs`|Each reads its learning rate from options rather than the framework default; TFC derives its spectrum bins from the actual RFFT output.|
|**Shared bases** <br> `src/Finance/Base/ForecastingModelBase.cs`, `src/Finance/Base/CrossSectionalGraphModelBase.cs`|A new cross-sectional graph base and expanded forecasting base absorb the per-model duplication.|
|**Portfolio** <br> `src/Finance/Portfolio/*`|Graph-attention and signature-transformer portfolio components move onto the shared base.|

**Estimated code review effort:** 4 (High) | ~75 minutes

**Possibly related PRs**

- [ooples/AiDotNet#1789](https://github.com/ooples/AiDotNet/pull/1789) — This is slice 16 of that PR, split for reviewable size.
- [ooples/AiDotNet#1984](https://github.com/ooples/AiDotNet/pull/1984) — Adds the AIDN070-076 analyzers that enforce the mechanical half of these same rules at build time.

**Poem**

> Vocoders learn the rate the paper set,
> Phase no longer squared, nor wrapped, nor met
> With optimizers built and left to idle —
> Now every step runs on a configured bridle.
> But count the heads before the tensors flow:
> Two was the number, eight is what they show.

src/TextToSpeech/Vocoders/PriorGradOptions.cs
ooples: Fixed:

```csharp
// PriorGrad's attention stack is 2 heads, and this has to be stated HERE rather than left
// to the base. PriorGrad.InitializeLayers used to pass a hardcoded 2 to
// CreateDefaultDiffusionVocoderLayers; replacing that with _options.NumHeads made the value
// configurable, but nothing assigned it, so it silently inherited TtsModelOptions.NumHeads =
// 8 and every default-constructed PriorGrad got a 4x wider attention stack than before.
NumHeads = 2;
```

Confirmed the chain rather than taking it on faith: `TtsModelOptions.cs:65` declares `public int NumHeads { get; set; } = 8`, `PriorGradOptions()` assigned every other field but not this one, and `git log -S "_options.NumHeads"` dates the substitution to this PR's own `1d614d4bf`.

Your framing is the part I recorded in the comment — this is not a refactor but a change to the *trained* architecture, so an existing checkpoint no longer matches the layer set it was trained into.

Setting the value in the constructor rather than overriding the property default, so it sits with the other paper values (`SampleRate`, `MelChannels`, `HopSize`, `NumDiffusionSteps`) and the copy constructor inherited from the base carries it unchanged.

Build: 4 `CS0246` before the change and 4 after, both the pre-existing `HippoMemoryCellLayer` / `Rwkv7Stack` cross-slice artifacts — this adds none.

PR #1991, opened 2026-08-08
fix(build): add the base hooks the #1789 split slices override

ooples:
## What

Adds the two base-class members that every `#1789` split slice overrides but that only ever existed on the `#1789` branch itself:

| Member | Where | Why the slices need it |
|---|---|---|
| `LayerBase<T>.ForwardTraced(Tensor<T>)` | `src/NeuralNetworks/Layers/LayerBase.cs` | ~16 layers across the TimeSeries slices declare `protected override Tensor<T> ForwardTraced(...)` |
| `FinancialModelBase<T>.TrainingOptimizer` | `src/Finance/Base/FinancialModelBase.cs` | `FactorVAE<T>` and siblings declare `protected override ... TrainingOptimizer` |

Plus `src/NeuralNetworks/Graph/LayerForwardObserver.cs`, which `Forward` now consults.

## Why standalone

This is the shared failure. Slices #1967–#1983 each target `master` independently, so every one of them hits the same `CS0115: no suitable method found to override` until #1789 lands in full. Landing just the base declarations unblocks all of them at once, in any merge order.

## The `Forward` change

`LayerBase<T>.Forward` goes from `abstract` to `virtual`. It now delegates to `ForwardTraced` and records the call when an observer is attached — the same `__call__`/`forward` split PyTorch uses, so the framework has one place to stand between the caller and the computation. **Existing layers that override `Forward` are unaffected**; they keep working and are simply invisible to tracing.

`ForwardTraced`'s default throws `NotSupportedException` rather than `NotImplementedException`: a layer overriding neither has no computation at all, and is not going to acquire one later.

## Verification

```
dotnet build src/AiDotNet.csproj -f net8.0   -> 0 Error(s)
dotnet build src/AiDotNet.csproj -f net471   -> 0 Error(s)
```

### DIFF
--- a/src/Finance/Base/FinancialModelBase.cs
+++ b/src/Finance/Base/FinancialModelBase.cs
@@ -44,6 +44,17 @@ namespace AiDotNet.Finance.Base;
 /// </remarks>
 public abstract class FinancialModelBase<T> : NeuralNetworkBase<T>, IFinancialModel<T>
 {
+    /// <summary>
+    /// Gets the model-specific optimizer used by the common tape training path.
+    /// A null value retains the neural-network default optimizer.
+    /// </summary>
+    /// <remarks>
+    /// A model that builds its own optimizer from its options - the paper's rate, its own beta values -
+    /// had no way to hand it to the shared <c>TrainWithTape</c> path, so the framework default silently
+    /// won and the configured optimizer never ran. Overriding this is that way.
+    /// </remarks>
+    protected virtual IGradientBasedOptimizer<T, Tensor<T>, Tensor<T>>? TrainingOptimizer => null;
+
     #region Execution Mode
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


ooples:
Measured on `origin/split/1789-17-finance-timeseries` (PR #1982), which is representative of the split slices — merged this branch into it and rebuilt `src/AiDotNet.csproj -f net8.0`:

| | errors | breakdown |
|---|---|---|
| before | **92** | 32 × CS0115, 30 × CS0534, 30 × CS0246 |
| after | **30** | 30 × CS0246 |

Every CS0115 (`no suitable method found to override`) and every CS0534 (`does not implement inherited abstract member Forward`) is gone. The 30 CS0246 that remain are the slice's own dependency on the option types in #1968 — a per-slice ordering issue, not a shared one.

PR #2010, opened 2026-08-16
feat(optimization): LP/QP/assignment solvers + fix solver-shaped defects

ooples:
﻿## Summary

Adds a first-class solver layer under `src/Solvers` and routes the library's existing hand-rolled optimization code through it. Every new solver has an Options class, XML docs citing the source paper, and integration tests checked against **known-optimal answers** rather than recorded output.

This started as "add the solvers the optimization literature needs" and grew as tracing internal consumers turned up a cluster of defects those solvers fix.

## New solvers

| Solver | Algorithm |
|---|---|
| `SimplexSolver` | Two-phase simplex, Bland's anti-cycling rule, dual values (Dantzig 1947) |
| `BranchAndBoundSolver` | Integer / mixed-integer programming (Land & Doig 1960) |
| `ActiveSetQuadraticProgramSolver` | Convex QP (Nocedal & Wright Alg. 16.3), reusing simplex for feasibility |
| `SequentialMinimalOptimizationSolver` | SVM dual, LIBSVM gradient form, maximal-violating-pair (Platt 1998) |
| `LinearAssignmentSolver` | Hungarian algorithm (Kuhn 1955, Munkres 1957) |
| `KktConditions` | Independent optimality certificate for constrained problems |

## General-function minimization

`GradientBasedOptimizerBase` now implements `IFunctionOptimizer<T>`, so every gradient optimizer gains `Minimize(x0, f, ?f)` by driving **its own existing** `Step(TapeStepContext<T>)` ? the function path and the training path are the same update code and cannot drift apart. A projected overload covers bound-constrained problems.

This removes three duplicate implementations the gap had caused:
- `Finance/Volatility/NelderMead.cs` ? deleted
- the hand-rolled Adam inside `GaussianProcesses/HyperparameterOptimizer`
- `LBFGSFunctionOptimizer` ? deleted; NOTEARS now uses `LBFGSOptimizer`

## Defects fixed

- **GradientEpisodicMemory** solved its dual QP with projected gradient descent; the paper solves it exactly with quadprog.
- **SuperLearner** ? comment said "active set method for NNLS", code was projected GD with a hardcoded learning rate that re-normalized weights *inside* the loop, so its fixed point was not the constrained optimum.
- **DETRSetLoss** matched greedily rather than with the Hungarian algorithm the paper requires. Separately, `ComputeTapeLoss` returned a **detached** zero tensor for non-structured input, so the loss reported non-zero while the gradient was identically zero ? training would silently never move.
- **QuantileRegression** ran OLS and returned before reaching its quantile code, fitting the conditional **mean** regardless of the requested quantile. Now the Koenker & Bassett (1978) linear program.
- **SupportVectorClassifier** used the simplified random-partner SMO.
- **LBFGSOptimizer** stored correction pairs violating the curvature condition, which the two-loop recursion divides by unguarded; now applies Powell damping. Curvature memory is cleared on every reset path, not only in `Optimize`.
- **NelderMeadOptimizer** compared raw fitness with hardcoded comparisons assuming a direction `IFitnessCalculator` need not have, and its **expansion branch was inverted**.
- **Five causal-discovery algorithms** defaulted to an unseeded secure RNG, so results were irreproducible run to run.
- **SupportVectorRegression** fitted OLS and returned ? no kernel, no ?-tube, no support vectors were ever computed.

## The `_useOLS` short-circuits ? all of them, now

Eleven models carried an unconditional or conditional `_useOLS = true` that made their real algorithm dead code. **All eleven are now gone.** Removing each one meant the documented algorithm ran for the first time, and in most cases that exposed further defects underneath ? those are fixed here too, not deferred.

### Unconditional short-circuits

`TimeSeriesRegression`, `SymbolicRegression`, `SupportVectorRegression`, `GeneralizedAdditiveModelRegression`, `NeuralNetworkRegression`, `MultilayerPerceptronRegression`, `LinearMixedModel`, `GeneralizedLinearMixedModel`, `DeepSurv`, `DeepHit`.

**`DeepSurv`** ? `TrainAsync` called `InitializeNetwork()` and then fitted least squares and returned. `ComputeCoxLossAndGradients` and `ComputeBaselineHazard` had no callers at all; `Epochs`, `BatchSize`, `LearningRate`, `L2Regularization`, `DropoutRate`, `UseBatchNormalization` and `EarlyStoppingPatience` were read by nothing. Now implements Katzman et al. (2018): mini-batch Adam with decoupled L2, dropout, batch normalization, and early stopping that restores the best epoch. Four defects surfaced:

- The **Cox gradient was not the derivative of the Cox loss**. It kept only the `i = k` term of `dL/dr_k = -d_k + exp(r_k) ? ?_{i: event, t_i ? t_k} 1/S_i`, and added it for censored subjects too where the sum runs over observed events alone.
- `ComputeBaselineHazard` exponentiated risk scores without the clamp the loss used; a large score overflowed to infinity, and the Breslow loop then subtracts that from itself, making every hazard value NaN.
- Features were fed in raw, so the linear predictor grew until `exp` overflowed.
- Batch-norm running statistics are parameters, and `Serialize` did not write them ? a restored network normalized with mean 0 / variance 1.

`Predict` returns expected survival time (integrating the survival curve, as lifelines' `predict_expectation` does), keeping it on the scale of the `y` passed to `Train`. `TrainAsync(x, times, events)` accepts censoring.

**`DeepHit`** ? same substitution; `ComputeLossAndGradients`, `SaveWeights` and `RestoreWeights` had no callers. Now implements Lee et al. (2018). Turning training on exposed gradients that had never run:

- The **ranking term's sign was inverted**. DeepHit eq. 8 uses `? = exp(-(F_i - F_j)/?)`, which shrinks as the ranking improves; the code formed `cifJ - cifI` and negated it, giving `exp(+(F_i - F_j)/?)` ? a penalty that *grows* when the ranking is right. It fought the log-likelihood term to a stalemate within a few epochs, so early stopping cut training off and 100 and 300 epochs returned identical predictions in identical time.
- **Two gradients were taken w.r.t. probabilities and used as if w.r.t. logits.** A correct softmax logit gradient sums to zero over cells; the censored term summed to `C/(1-C)` and the ranking term added a bare constant. Both injected a spurious uniform shift into every update.
- `ApplyLayer` dropped units on **every** call including `Predict`, so the same model returned a different answer each time.
- Features were raw, saturating the shared softmax.
- The time grid was sized by option alone: 100 bins against 100 subjects is ~1 event per bin, so most softmax cells got no signal and the predicted expectation collapsed to the mean observed time. It is now capped at what the sample supports.

### Conditional short-circuits

`LogisticRegression`, `MultinomialLogisticRegression` and `BetaRegression` detected a target their algorithm could not model and, rather than saying so, fitted least squares ? then reported that linear fit's coefficients as their own, with `Predict` skipping the sigmoid so a caller who asked for a probability received an unbounded number.

No mainstream library does this: scikit-learn raises through `check_classification_targets`, statsmodels `Logit` raises, R's `glm(family=binomial)` errors, and `betareg` rejects a target outside (0,1). Validation now lives in `ValidationHelper` beside `ValidatePoissonData`, whose comments already argued this doctrine, and the message names the observed target and the model to use instead.

Two targets that used to reach the least-squares path now train properly, so this is a capability gain and not only a new restriction:

- **Any two labels** are accepted, not only literal `{0,1}` ? `{-1,1}`, `{1,2}`, `{3,7}` are label-encoded as scikit-learn's `LabelEncoder` does. The old check demanded `{0,1}` exactly, so equivalent labels fell through to OLS.
- **Class labels need not be consecutive or zero-based**; `{2,5,9}` encodes to `{0,1,2}` and `Predict` maps back, so predictions return in the caller's own labels.

Three more defects came out with the fallbacks:

- `MultinomialLogisticRegression` **quantized** an integer target with many distinct labels into at most 5 equal-width bins, then predicted bin indices matching no training label.
- `BetaRegression` **min-max rescaled** a target already inside [0,1] onto (0.01, 0.99), stretching a genuine [0.3, 0.7] range across nearly the whole interval. Beta regression models a proportion directly (Ferrari & Cribari-Neto 2004).
- That rescaling was inverted only `if (_needsTransform)`, which was true only on the OLS path and therefore **always false on the beta path**. It was never undone, so `PredictAsync` and `PredictDistributionsAsync` both reported the distorted scale as the caller's.

Boundary proportions get a cited route rather than a rejection: `BetaRegressionOptions.CompressBoundaryValues` applies the Smithson & Verkuilen (2006) transformation `y' = (y(n-1) + 0.5)/n`. Opt-in, because moving a caller's data unasked is the same hidden substitution being removed.

### Three defects the fallbacks had been hiding

With the substitution gone, these models met the model-family harness for the first time. All three failures were in the models:

- **`LogisticRegression` learned the response backwards.** Batch gradient ascent moves the intercept by at most `LearningRate` per iteration, so on raw features it cannot reach the intercept the fit needs ? for features spanning [0,10] with a balanced split that is near **-60**, and at the default 0.01 over 1000 iterations it can travel about 5. The coefficients absorbed the shortfall by turning negative: measured `[-0.15, 0.03, 0.42]` where the data-generating coefficients were 2, 4, 6, so every prediction *fell* as the feature rose. It now fits on standardized features and transforms the coefficients back, giving `[0.13, 0.27, 0.57]` on the caller's own scale.
- **`MultinomialLogisticRegression` published a coefficient vector as long as the number of classes.** `_coefficients` is `[class, feature]`, so `GetColumn(0)` is "feature 0's weight for each class". With four classes over three features it reported a coefficient for feature 3 and `GetActiveFeatureIndices` returned an index outside the input feature space.
- **`BetaRegression`'s Fisher scoring was wrong in both terms.** `LinkFunctionDerivative` returns `g'(?) = d?/d?`, but the local was named `dmu` and both formulas used it as `d?/d?`. GLM Fisher scoring needs `w = 1/(V(?)g'(?)?)` and `z = ? + (y-?)g'(?)`; the code had `?V/g'?` (which for the logit is `?V?`) and `? + (y-?)g'/V`, an extra factor of `1/V`. The working response was fatal: at ? = 0.1 every step overshot more than tenfold, coefficients ran away, and the fitted mean saturated at 0 or 1 for every observation. Predictions on a target spanning [0.084, 0.907] came back `[1.0000, 1.0000, 0.9772, 0.0007, 0.0000]`; they are now `[0.31, 0.41, 0.52, 0.62, 0.72]`.

### Harness change

`RegressionModelTestBase` gained a `ToTarget` hook so a model whose response domain is narrower than "any real number" is exercised on data it can actually be fitted to, instead of opting out of the structural invariants. It defaults to passing the target through unchanged, so **no other model is affected**. Logistic thresholds at the median, multinomial bins into four ordered classes, and Beta maps through a logistic squash into (0,1) ? each order-preserving, so the invariants still mean what they meant. All three now run the full contract: output dimension, determinism, cloning, serialization, metadata, collinear features, single feature, active feature indices.

`DeepSurv` and `DeepHit` declare the identity-link invariants inapplicable, joining the six existing non-identity-link regressions (Poisson, Gamma, Inverse Gaussian, Negative Binomial, Tweedie, Beta). Cox has a log-hazard link and no intercept, and DeepHit predicts an expectation over a bounded discrete grid. Those invariants passed only because the models **were** least squares, which is an identity-link model ? they were measuring the substitution. Eleven survival-specific tests replace them: oracle-anchored concordance, concordance under 35% censoring, invariance of the risk ordering to a monotone transform of time, PMF summing to one, prediction determinism under dropout, cause separation across two competing risks, serialization round trips, and input validation.

## Verification

Targeted suites, all measured:

- `FunctionMinimizationIntegrationTests` ? **34/34** (13 optimizers on sphere, ill-conditioned quadratic, Rosenbrock via L-BFGS, box projection, derivative-free Nelder-Mead)
- `IntegrationTests.Solvers` ? **63/63** (hand-solved LP vertices, strong duality, complementary slackness, 0/1 knapsack, KKT-certified QP, Hungarian vs brute force)
- SVM ? **139/139**; SVR ? **78/78** with genuine kernel SVR running for the first time
- Detection ? **556/556**, including a pre-existing `DETRSetLoss_Gradient_IsNonZero` failure now fixed
- SuperLearner + GEM + ContinualLearning ? **329/329**
- `QuantileRegressionLinearProgramTests` ? **9/9**
- `CausalDiscoveryDeterminismTests` ? **4/4**
- `DeepSurvTests` ? **30/30**; `DeepHitTests` ? **30/30**
- `LogisticRegressionTests` + `MultinomialLogisticRegressionTests` + `BetaRegressionTests` + `GLMFamilyRegressionIntegrationTests` ? **107/107**, where 55 were failing when the fallbacks were first removed
- `SymbolicRegressionTests` + `GeneralizedLinearMixedModelTests` ? **45/45**
- `AdvancedRegressionIntegrationTests` ? **15/15**; `RemainingRegressionIntegrationTests` ? **34/34**

Builds clean on net10.0 and net471.

## Notes for review

- `SymbolicRegression` and `GeneralizedLinearMixedModel` were listed as deliberately-failing in an earlier revision of this description. Both were fixed later on the branch (`35e4bec97`, `57adfa141`) and their 45 tests pass; that section is obsolete and has been removed.
- Two CI checks are red for reasons outside this branch: `Model Performance Coverage and Regressions` also fails on `master`, and `commitlint` rejects eight commit subjects that landed without the local `.githooks/commit-msg` hook. Fixing the latter needs a history rewrite and force-push, which `commitlint-fix.yml` deliberately does not do on an active PR.

?? Generated with [Claude Code](https://claude.com/claude-code)

## Commit-history validation

The PR's commitlint failure was repaired without changing the implementation. Nineteen non-compliant subjects in the branch history were rewritten to the repository's conventional-commit types (including the two `merge:` subjects visible in the failed job).

Evidence on the rewritten head:

    npx commitlint --from origin/master --to HEAD --config commitlint.config.mjs
    0 errors

    old head tree: fdbd430a2d412c24ccfcbf02ce7b941777375859
    new head tree: fdbd430a2d412c24ccfcbf02ce7b941777375859

The identical tree IDs prove this was metadata-only: no source, generated code, tests, or runtime behavior changed.


ooples:
Both blocking failures are resolved, so this is out of draft.

**SymbolicRegression** — the OLS short-circuit was hiding four real defects, all now fixed:
- `GetSubMatrix` takes `(startRow, startColumn, rowCount, columnCount)`, but the split size was passed as the **start column**, so every train/val/test split had zero rows while its target vector kept full length. Same bug in `GeneticAlgorithmRegression`.
- No intercept: the evolved expression was forced through the origin, so translation equivariance was impossible.
- `GetActiveFeatureIndices` fell through to the kernel-model base, which reads `Alphas`/`SupportVectors` this model never populates.
- `VectorModel.TrainInternal` threw on a singular `XᵀX`, making it unusable as a GA population member; now ridge-regularises, as its own error message advised.

**GeneralizedLinearMixedModel** — two separate problems:
- Model: `ExtractFixedEffectsMatrix` never added a constant column, so the model had no intercept.
- Harness: GLMM defaults to Binomial/logit (a probability in (0,1)), but the regression base asserts continuous-response invariants — the inverse logit saturates at 1.0 and R² collapses. The test now uses the Gaussian identity-link family those invariants actually apply to.

Also fixes a regression introduced earlier in this branch: the SMO refactor computed step curvature from the label-signed `Q` rather than the raw kernel, which cancels to exactly zero for an opposite-label pair.

Verified 93/93 across MixedModel, SymbolicRegression, GeneticAlgorithmRegression, IntegrationTests.Solvers and SupportVector.

Still outstanding on this branch: ~7 further models carry the same `_useOLS` short-circuit, and A3/A4 (interior point, augmented Lagrangian, control module) are not yet built.

ooples:
### Follow-up: the function-minimization API was reachable on exactly one optimizer

Writing course material against `Minimize` surfaced two defects in the A1 rollout.

**1. `Minimize` was unreachable on 27 of the 28 optimizers that had it.** It is implemented on
`GradientBasedOptimizerBase`, so every gradient optimizer inherited it — but only `LBFGSOptimizer`
had `CreateForFunction`, and every other constructor demands an `IFullModel` that a plain function
does not have. Each of the 27 now has the same factory, with a private options-only constructor
where the public one takes a non-nullable model.

`NelderMeadOptimizer` was left alone: its public options-only constructor is not ambiguous with the
model constructor (that one has no default for `model`), so there was nothing to fix, and changing
it would have broken four call sites for no benefit.

**2. L-BFGS converged linearly rather than superlinearly.** Chained Rosenbrock to a gradient
tolerance of 1e-6:

| n | before | after |
|---|---|---|
| 2 | 253 | 64 |
| 5 | 7997 | 75 |
| 10 | 8379 | 101 |
| 20 | — | 183 |

Diagnosed by sweeping `PowellDampingFactor` rather than by reading the code: at 0 the same problems
took 107 and 163 evaluations at n = 5 and 10, so damping was doing the damage. It was doing it
because the line search only enforced sufficient decrease. Nocedal and Wright's L-BFGS assumes a
step satisfying the Wolfe conditions, whose curvature half is exactly what makes `sᵀy > 0` hold;
without it every pair needed Powell's repair, and the repair distorts the curvature information the
two-loop recursion runs on.

`Minimize` now runs the strong Wolfe search (Algorithms 3.5 and 3.6 — bracketing plus zoom), skips
damping for pairs from a step that satisfied the curvature condition, and carries the accepted
point's evaluation into the next iteration so the search's last trial is free rather than repeated.
Backtracking remains as the fallback and as an opt-out via `UseStrongWolfeLineSearch`;
`WolfeCurvatureConstant`, `LineSearchMaxZoomSteps` and `LineSearchMaxStep` are configurable
alongside it.

Also fixes four uses of `double.IsFinite` in test files, which does not exist on net471 and was
breaking that leg of the build.

**Verification.** 489/489 green across `FunctionMinimization`, `Metaheuristic`,
`OptimizerConvergenceCheckPattern`, NOTEARS, GP hyperparameters and Finance volatility — the suites
that exercise `Minimize` and the three migrated call sites. 32/32 green on the new
`FunctionOptimizerFactoryIntegrationTests`, which assert the evaluation budgets above so a
regression fails loudly rather than quietly. net471 and net10.0 both build clean.

One thing I could not verify cleanly: a run of `--filter "FullyQualifiedName~Optimizers"` produced
54 failures, all of them `OptimizerStep_ParamL2_DoesNotExplode` across model-family shards, and all
reporting "Test execution timed out after 120000 milliseconds" while the machine was under disk
pressure. That test is a neural-network training invariant on the Adam path and does not touch
`Minimize` or `UpdateLBFGSMemory`'s changed signature, but I am reporting it rather than asserting
it is unrelated.

ooples:
### A22: the same line-search gap on `BFGSOptimizer`, `DFPOptimizer` and `ConjugateGradientOptimizer`

Found while writing course material against the function-minimization API, and **not fixed in this
PR** — recording it so it is not lost.

`BFGSOptimizer`, `DFPOptimizer` and `ConjugateGradientOptimizer` have no `Minimize` override, so
they inherit `GradientBasedOptimizerBase`'s one-`Step`-per-iteration loop with **no line search at
all**. Their `LineSearch` exists only on the model-training `Optimize` path. This is the same
defect the strong-Wolfe commit fixed for L-BFGS.

Measured on chained Rosenbrock from the usual hard start, gradient tolerance `1e-8`, counting
objective evaluations:

| variables | BFGSOptimizer | LBFGSOptimizer |
|---|---|---|
| 2 | 51 | 65 |
| 5 | **1911** | 84 |
| 10 | **1166** | 113 |
| 20 | **673** | 169 |

BFGS wins at n=2 and then loses by more than twenty-fold — which is the wrong shape for a method
that keeps the *whole* curvature history against one that keeps ten pairs.

**Why it is not in this PR.** The clean fix is to lift the line-search machinery out of
`LBFGSOptimizer` into `GradientBasedOptimizerBase` as protected members, move the eight
line-search options from `LBFGSOptimizerOptions` up to `GradientBasedOptimizerOptions`, and give
each of the three a `Minimize` override supplying its own direction rule. That is a real change to
two 30 KB files plus three new direction implementations, and this PR is already large. I started
it, judged the automated edit too risky to do safely at this point, and backed it out rather than
half-land it.

Nothing regressed — before the `CreateForFunction` commit these three optimizers had no reachable
`Minimize` at all, so this is an unfinished feature rather than a break.

The course lesson on quasi-Newton methods (2.7) therefore compares its own hand-written BFGS
rather than the library's, and says so in the text.

ooples:
### A22 landed: the Wolfe line search is now shared, and BFGS/DFP/CG use it

Follow-up to the previous comment — the fix is in, done by hand rather than by the automated edit I
backed out.

**What moved.** The line-search machinery and the eight options that govern it now live on
`GradientBasedOptimizerBase` and `GradientBasedOptimizerOptions`, where they always belonged: every
line search wants `ArmijoConstant`, `WolfeCurvatureConstant`, `UseStrongWolfeLineSearch` and the
rest, not just L-BFGS's. `MemorySize`, `MinimumCurvature` and `PowellDampingFactor` stay on
`LBFGSOptimizerOptions`, being genuinely L-BFGS-specific. A shared `MinimizeWithLineSearch` drives
the loop and each method supplies only its own direction rule, so the three new implementations are
about forty lines each rather than three copies of the loop.

**Chained Rosenbrock, gradient tolerance 1e-8, objective evaluations:**

| n | BFGS before | BFGS after | DFP | CG before | CG after | L-BFGS |
|---|---|---|---|---|---|---|
| 2 | 51 | **65** | 59 | — | **379** | 65 |
| 5 | 1911 | **65** | 2868 | — | **2204** | 84 |
| 10 | 1166 | **100** | 20096 | — | **2563** | 113 |
| 20 | 673 | **145** | 20128 | — | **3125** | 169 |
| 50 | — | **311** | 20261 | — | **5761** | 372 |

BFGS now beats L-BFGS at every size above two, which is the ordering theory predicts since it keeps
the whole curvature history rather than ten pairs.

**Three things the measurements decided rather than the design.**

Conjugate gradient first came out at **165 066** evaluations — worse than useless. The recursion was
recording the gradient at the *new* point as "previous", so every β was zero and the method was
steepest descent wearing a hat. It also has to remember the *direction* rather than the step, since
the step folds in the accepted length and rescales the memory term arbitrarily.

`ConjugateGradientOptimizerOptions` now defaults `WolfeCurvatureConstant` to **0.1** rather than the
inherited 0.9. Nonlinear CG only produces a descent direction when the search is accurate, and N&W
recommend exactly this split (§5.2).

BFGS scales its initial estimate by `(sᵀy)/(yᵀy)` before the first update (N&W eq. 6.20), taking it
from 97 to 65 evaluations at five variables. **The same change made DFP strictly worse** — 20132
everywhere, converging nowhere — so it is applied to BFGS only, on the measurement rather than on
the symmetry.

DFP remains the weakest: it solves 2 and 5 variables and exhausts its budget beyond that. That is
the documented result which made BFGS the default, and it is still an improvement on having no line
search at all.

**Verification.** 521/521 green across `FunctionMinimization`, `Metaheuristic`,
`OptimizerConvergenceCheckPattern`, NOTEARS, GP hyperparameters, Finance volatility and the factory
tests. 45/45 on `FunctionOptimizerFactory`, including new budget assertions for all three methods so
a regression fails loudly. net471 and net10.0 both build clean.

ooples:
### A20 — the derivative-free search methods can now actually minimize a function

`IDerivativeFreeFunctionOptimizer<T>` had exactly one implementer, `NelderMeadOptimizer`. Eight other optimizers in `src/Optimizers/` advertised a search algorithm with no way to run it on a plain function, and their public constructors all demand an `IFullModel` that a plain function does not have.

All eight now implement the interface against their published algorithm, each with a `CreateForFunction` factory:

| Optimizer | Paper |
|---|---|
| `SimulatedAnnealingOptimizer` | Kirkpatrick, Gelatt and Vecchi 1983 |
| `ParticleSwarmOptimizer` | Kennedy and Eberhart 1995, Clerc and Kennedy 2002 constriction |
| `DifferentialEvolutionOptimizer` | Storn and Price 1997, DE/rand/1/bin |
| `GeneticAlgorithmOptimizer` | real-coded, blend crossover, elitism |
| `CMAESOptimizer` | Hansen and Ostermeier 2001, Hansen 2016 constants |
| `PowellOptimizer` | Powell 1964, with Brent's direction test (NR §10.5) |
| `TabuSearchOptimizer` | Glover 1989, with aspiration |
| `AntColonyOptimizer` | Socha and Dorigo 2008, ACO-R |

Shared machinery lands on `OptimizerBase`: `DerivativeFreeSearch` (which tracks the best point *ever seen* — several of these methods deliberately accept worse points, so "where it ended" is the wrong answer to return), `ValidateMinimizeArguments`, `CreateSearchRandom` seeded from `Options.Seed`, and `NextGaussian`.

#### Measured, not asserted

3-variable sphere from `(10, -8, 4)`, minimum 0 at the origin:

| method | evaluations | final f |
|---|---:|---:|
| NelderMead | 300 | 1e-20 |
| SimulatedAnnealing | 2001 | 2.17e-4 |
| ParticleSwarm | 17451 | 0 |
| DifferentialEvolution | 9701 | 0 |
| GeneticAlgorithm | 27821 | 6.44e-7 |
| CMA-ES | 2823 | 0 |
| Powell | 578 | 0 |
| TabuSearch | 10721 | 0 |
| AntColony | 2201 | 0 |

Rosenbrock from `(-1.2, 1)`, minimum 0 at `(1, 1)`:

| method | evaluations | final f |
|---|---:|---:|
| NelderMead | 243 | 0 |
| SimulatedAnnealing | 2001 | 1.31e-3 |
| ParticleSwarm | 49901 | 0 |
| DifferentialEvolution | 11901 | 0 |
| GeneticAlgorithm | 198101 | 5.92e-6 |
| CMA-ES | 34001 | 0 |
| Powell | 2259 | 0 |
| TabuSearch | 10841 | 8.1e-19 |
| AntColony | 18301 | 0 |

Simulated annealing and the genetic algorithm stopping a few digits short is not a defect — they explore rather than polish, and the test that demands machine precision deliberately excludes them. All nine are reproducible from a fixed `Options.Seed`.

#### Two bugs found by measuring rather than reading

- **Powell stalled at f = 3.98 on Rosenbrock.** When both probes along a direction came back uphill, the bracket excluded the interior, so the line search could never take a step *shorter* than its initial probe — in a narrow curved valley that means it cannot move at all.
- **Ant colony sat at 115.1 on the sphere.** The sampling width was being taken from `InitialPheromoneEvaporationRate`, which is a different quantity from ACO-R's width scale despite the shared vocabulary.

#### Verification

```
Passed!  - Failed: 0, Passed: 62, Skipped: 0, Total: 62   (DerivativeFreeFunctionOptimizerIntegrationTests)
Passed!  - Failed: 0, Passed: 745, Skipped: 0, Total: 745 (IntegrationTests.Optimizers + IntegrationTests.Solvers + UnitTests.Optimizers)
```

Both target frameworks build clean, net471 included.

Commit: `8a507ec37`

ooples:
### A23–A25 — three defects found by writing Chapter 4 against the library

**A23 · `SvdDecomposition.Solve` compared singular values against exact zero.** A rank-deficient matrix's mathematically-zero singular value arrives from a floating-point computation as `4.93e-32`, so it was inverted. On a 4×3 matrix whose third column is twice its second:

| | before | after |
|---|---:|---:|
| ‖x‖ | 9.007e+15 | 1.0954 |
| sum of squared residuals | **642** | 4.1e-30 |

The correct minimum-norm answer is `(1, 0.2, 0.4)`. Now uses the LAPACK `gelsd` / SciPy `lstsq` convention — `max(m,n)·eps` relative to the largest singular value — exposed as a constructor parameter alongside new `RelativeTolerance` and `Rank` properties. Passing `0` restores the previous behaviour for anyone who wants an exact-arithmetic pseudoinverse. **10 new integration tests.**

**A24 · four model defects, all pre-existing.** I checked first whether A23 caused them — identical 9 failures with and without it.

- **`LinearMixedModel` had no fixed intercept.** The design was built from raw non-grouping columns and `Intercept` was set to zero, so the response's overall level had nowhere to live but the random effects — which are shrunk toward zero and contribute *nothing* for an unseen group. Shifting every target by 1000 moved predictions by 1530; now 1000.00. Its `Clone` also fell through to `base.Clone()` and predicted by throwing `"Model must be trained"`.
- **GLMM's BLUPs were not constrained mean-zero**, so the alternating scheme settled at a fixed point splitting the level: a shift of 1000 moved the intercept by 499.99. Recentring gives 1000.00. Measured against `LinearMixedModel` side by side — that comparison is what proved the invariant satisfiable rather than inapplicable.
- **`MultilayerPerceptronRegression` trained on the raw response**, so its output layer could not reach 1000 from a small random initialization within its epoch budget: scaling targets by 100 scaled predictions by 5.16. Now standardizes the target and inverts at prediction. `GetActiveFeatureIndices` inspected a linear coefficient vector a perceptron does not have. Weight initialization was unseeded — and my first fix used `string.GetHashCode()`, which .NET randomizes *per process*, so it stayed flaky until a fixed fallback.
- **`CenterAndScale_SingleRow_ReturnsNaN`** asserted that standardizing a constant column produces NaN. The helper was repaired earlier to scale zero variance by one, as scikit-learn's `StandardScaler` does; the test still pinned the defect.

Two shared-suite hooks, following the existing `IdentityLinkInvariantsApplicable` precedent: `StructuralFeatureIndices` (a mixed model's grouping column is a factor, as lme4 treats `y ~ x + (1|group)`) and `CoefficientOrderingInvariantApplicable`. I first tried making the probe point representative instead — it fixed the perceptron and broke four tree models, so it was reverted.

**A25 · 32 test failures that were not model defects at all.** Six families were failing, a different subset each run.

| | result |
|---|---|
| each of the six families alone | **22/22 passed** |
| full filter, parallel collections | 32 failed / 1245 passed |
| full filter, serial collections | **0 failed / 1277 passed** |

`TensorArena.Create` activates an arena per thread, but arenas pool their backing arrays into a **cross-arena persistent pool** on `Dispose`, and that pool is shared process-wide — so tests holding arenas on different threads hand each other arrays. `TensorArena` lives in the `AiDotNet.Tensors` package, so **the thread-safety fix belongs there and cannot be made from this repository**; until it lands, running these collections concurrently is not valid. `xunit.runner.json` now says so, with the measurements. Cost: 10m14s serial against 9m36s parallel — these tests are compute-bound and already saturate the machine.

#### Verification

```
ModelFamilyTests.Regression        1277 passed,  0 failed
SvdRankDeficientSolveTests           10 passed,  0 failed
LinearMixedModel + RegressionHelper  61 passed,  0 failed  (was 5 failed)
MultilayerPerceptron                 22 passed,  0 failed  (was 3 failed)
CoefficientSigns, every family       61 passed,  0 failed
```

Commits: `65b2d5b74`, `68e2fdb1c`

ooples:
Commitlint is repaired on head `24af0635b`. I audited the complete PR range rather than only the final two errors shown in the truncated job log: all 19 non-compliant subjects are now conventional, and the exact local commitlint invocation exits 0. Old and new heads have the identical tree `fdbd430a2d412c24ccfcbf02ce7b941777375859`, proving the history rewrite changed no implementation or test content. The PR description now records this evidence. @coderabbitai review

ooples:
### Readiness repair after latest-master sync

Updated head: `97a8a7641`, based on current master `cace89f61` (#2026).

The two serving regressions were caused by restored registered attention weights leaving derived projection caches stale. The fix centralizes parameter-change notification in `LayerBase`, invalidates persistent tensors during in-place restore, and makes `PagedCachedMultiHeadAttention` clear its transposed/quantized/GEMM weight caches. Regression coverage now checks batched-vs-independent decode and restore-after-warmup behavior.

Local validation on the current baseline:
- `net10.0` build: 0 warnings, 0 errors
- `net471` build: 0 errors (existing project warnings remain)
- cache restore/registration tests: 5/5
- full serving suite: 289/289
- solver/DETR/generator slice: 288/288
- full PR-range `git diff --check`: clean
- exact repository commitlint command: exit 0

The accidental update introduced seven non-conventional subjects; only those subjects were rewritten, with a recovery ref retained and source-tree identity verified across the message rewrite. The two actual code commits remain conventional.

@coderabbitai review

PR #2030, opened 2026-08-19
feat(adamw): let a recipe exempt parameters from decoupled weight decay

ooples:
Follow-up to #2029, which needed this and could not have it.

## Why

`WeightDecay` is a single coefficient applied to the **whole flat parameter vector**, so there is no way to express *"decay these parameters but not those"*. Published recipes routinely need exactly that:

- **RecurrentGemma** (Botev et al., 2024) §2: *"we do not apply weight decay to the parameters of the recurrent (RG-LRU) layers during training."*
- The ordinary transformer recipe exempts biases and normalization gains.

The reason is the same in both cases: decay pulls weights toward zero, which helps where magnitude is what's being regularized, and corrupts parameters whose **value** carries meaning — a recurrence's own decay rate, for instance.

In #2029 I left this measure out rather than ship a property that silently did nothing. This is the capability that makes it expressible.

## What changed

`WeightDecayMask` — an optional per-parameter multiplier on the decay term.

- **`null` decays everything**, which is exactly what every caller had before. No existing model changes behaviour.
- Entries multiply elementwise: `1` decays normally, `0` exempts, fractions scale.
- The gradient update is untouched — this is decoupled decay only.
- Applied at **all three** eager sites (two vector paths and the element-wise span loop) so they cannot disagree.

## The fused path

`TryGetFusedOptimizerConfig` now **declines** when a mask is set.

The fused config carries decay as a single `float`, so a masked run cannot be represented in it — the compiled kernel would keep decaying the parameters the eager path exempts. Declining is the same mechanism that method already uses for adaptive learning rates.

Silently diverging between the compiled and eager paths would be the worst option available: it would surface only as a slow accuracy drift, with nothing pointing at the optimizer.

## Verification

| Test | Asserts |
|---|---|
| `NullMask_DecaysEveryParameter...` | the default is unchanged behaviour |
| `ZeroMaskEntries_ExemptExactlyThoseParameters` | exemption reaches the update |
| `FractionalMaskEntries_ScaleTheDecay` | 0.5 mask produces exactly half the movement |
| `MaskPresent_DeclinesFusedCompilation` | masked declines, unmasked still fuses |

Each isolates the decay term by stepping with a **zero gradient**, so anything that moves a parameter is decay and only decay.

**75 optimizer tests pass unchanged**, including `FusedSpecMatchesEagerBehaviour` and the copy-constructor suite — this touches an optimizer every model uses, so that regression check is the point.

## Note for reviewers

This is shared-optimizer code. The two things that make it safe are the `null` default (byte-identical behaviour for every current caller) and the fusion opt-out, and both are directly tested rather than asserted.

A natural next step, not included here: have `RecurrentGemmaLanguageModel` build a mask that zeroes its `RealGatedLinearRecurrenceLayer` slots, which would complete the paper's third measure. That needs a layer-to-flat-slot mapping and is worth its own PR.

### DIFF
--- a/src/Models/Options/AdamWOptimizerOptions.cs
+++ b/src/Models/Options/AdamWOptimizerOptions.cs
@@ -142,6 +142,36 @@ public class AdamWOptimizerOptions<T, TInput, TOutput> : GradientBasedOptimizerO
     public double WeightDecay { get; set; } = 0.01;
 
     /// <summary>
+    /// Gets or sets a per-parameter multiplier applied to the decay term, or <c>null</c> to decay
+    /// every parameter equally.
+    /// </summary>
+    /// <value>A vector as long as the flat parameter vector, or <c>null</c>. Defaults to <c>null</c>.</value>
+    /// <remarks>
+    /// <para>
+    /// <see cref="WeightDecay"/> is a single coefficient applied to the whole flat parameter vector, so
+    /// there is otherwise no way to express "decay these parameters but not those". Published recipes
+    /// routinely need exactly that: RecurrentGemma (Botev et al., 2024) Section 2 states "we do not
+    /// apply weight decay to the parameters of the recurrent (RG-LRU) layers during training", and the
+    /// usual transformer recipe exempts biases and normalization gains.
+    /// </para>
+    /// <para>
+    /// Entries multiply the decay term elementwise: 1 decays normally, 0 exempts, and values in
+    /// between scale it. The gradient update is untouched -- this is decoupled decay only.
+    /// </para>
+    /// <para>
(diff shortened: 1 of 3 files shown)


ooples:
Superseded by #2029. `feat(adamw): let a recipe exempt parameters from decoupled weight decay` has been merged into `fix/n1-nonfinite-training`, along with the rest of the N1 cluster work.

This PR described itself as a follow-up to #2029, so keeping it on a separate branch just split one review across two places. The `WeightDecayMask` option, the fused-config opt-out, and `AdamWWeightDecayMaskTests` are all present on #2029 unchanged.

Closing here; review continues on #2029.

PR #2033, opened 2026-08-20
fix(n1): repair non-finite quantum state encoding and mbpa retrieval

ooples:
Fixes pre-existing **N1 (non-finite training)** CI failures. Four product-code defects, one
test-side target projection. No assertion, tolerance or threshold was changed.

## QuantumNeuralNetwork — 10 failures -> 0

**`PrepareQuantumState` took `Sqrt` of the raw feature value.** That is only defined when every
feature is non-negative and already sums to one. The fixture input carries 71 negative values out
of 128 (min -0.99), so `Sqrt` returned NaN and that NaN flowed through every layer into the
Born-rule measurement. It also produced states whose probabilities did not sum to one, which is not
a valid quantum state. Replaced with the standard construction — normalise to unit L2 norm and use
the result as amplitudes, as Qiskit `initialize` and PennyLane `AmplitudeEmbedding` do. Total on
real inputs, keeps the sign `Sqrt` discarded, guarantees `sum |psi|^2 == 1`. Zero-norm falls back to
uniform superposition instead of dividing by zero.

**`QuantumLayer` normalised by `sum(|state|^2) + eps` instead of its square root.** Its own comment
says `sqrt(...)` and its GPU path does exactly that ("Step 4: Sqrt to get L2 norm") — so the CPU and
GPU paths disagreed on the state convention. Dividing by the SQUARED norm leaves a state of length
`1/||state||`: at the default 128-feature input roughly a 128x shrink, applied at each of two
quantum layers, so `Predict` returned ~4.6e-4 and gradients underflowed to zero.

## MbPA — 3 failures -> 0

**`MbPAEpisodicMemory.Retrieve` returned neighbours in max-heap order.** The bounded heap keeps the
WORST kept candidate at the root, so `kept[0]` was the FARTHEST of the k selected. A method
contracted to return "the k nearest" handed them back worst-first, and the kernel weight belonging
to the nearest entry was reported against the farthest (0.25 where 0.75 was correct). The kernel
`1 / (eps + d^2)` and its normalisation were already correct — only the order was wrong.

**`MbPAAdaptedModel.AssembleOutput` threw for a batched multi-component head**, so a model with
`OutputDimension > 1` could not predict a batch at all. The throw was an over-correction to a real
earlier bug (silent truncation to component 0); `rows * outputDim` values fit a flat `Vector<T>`
losslessly and are now written row-major, matching how the `Matrix<T>` branch already packs them.
This cannot regress a caller — the path it replaces threw.

## Born-rule target projection (test-side)

`LossStrictlyDecreasesOnMemorizationTask` handed a Born-rule head a NEGATIVE target, which |psi|^2
cannot represent at any parameter value. Measured: target -0.155572, prediction driven 0.0517 ->
0.0053 (the model correctly walking toward 0, the nearest reachable point), loss pinned at
0.024203 == (-0.155572)^2 — the entire residual is the unreachable sign. The invariant read a
converged model as a learning failure. `MakeTargetWellPosedForLoss` already performs this projection
for `CrossEntropyWithLogitsLoss` heads for the same reason; this adds the Born-rule case. Scope is
provably one model: `BornRuleMseLoss` is the default loss of `QuantumNeuralNetwork` and nothing else
in `src`.

## Verification

| Suite | Before | After |
|---|---|---|
| QuantumNeuralNetworkTests | 10 failed | **29 passed, 1 skipped, 0 failed** |
| MbPA (all) | 3 failed | **20 passed, 0 failed** |
| MetaLearning sweep | — | 481/482 (remaining failure `MetaLearnerBase_AccessorsAndFallback_AreCovered` is pre-existing and unrelated — a reflection `TargetParameterCountException` in a coverage test) |

Draft: further N1-cluster families still to come on this branch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Batched vector predictions with multiple output components now provide clearer guidance on using matrix or tensor outputs.
  - Nearest-neighbor results are ordered from closest to farthest, with matching kernel weights.
  - Quantum models now handle signed, extreme, tiny, and zero inputs safely through stable normalization.
  - Probability-based targets are normalized correctly, including uniform handling for all-zero targets.
  - Improved numerical stability for sequence anomaly detection and video-memory attention.
  - Stateful model comparisons now reset state correctly and handle matching non-finite values.

### DIFF
--- a/src/MetaLearning/Algorithms/MbPAAdaptedModel.cs
+++ b/src/MetaLearning/Algorithms/MbPAAdaptedModel.cs
@@ -189,29 +189,14 @@ public partial class MbPAAdaptedModel<T, TInput, TOutput> : MetaLearningModelBas
 
             if (rows.Count == 1) return (TOutput)(object)rows[0];
 
-            // BATCHED MULTI-COMPONENT: FLATTEN, DO NOT REFUSE. The concern that motivated the throw
-            // was TRUNCATION -- the old code kept component 0 of each row and silently dropped the
-            // rest. Refusing fixed that, but it over-corrected: a flat Vector<T> holds
-            // rows.Count * outputDim values perfectly well, so writing the rows out consecutively
-            // discards nothing. The result is ordered row-major (example 0's components, then
-            // example 1's, ...) and OutputDimension is what a caller reshapes by, exactly as it is
-            // for the Matrix<T> branch below.
-            //
-            // This cannot regress an existing caller: the combination it replaces THREW, so no
-            // working code path could have depended on the previous behaviour. What it does fix is a
-            // model configured with a multi-component head being unable to predict a batch at all
-            // (MbPAMechanismTests.AdaptedModel_PredictsWithoutRetainingAnyAdaptation predicts a
-            // 4-row query through a 3-component head and never reached its transience assertions).
-            var flattened = new Vector<T>(rows.Count * outputDim);
-            for (int i = 0; i < rows.Count; i++)
-            {
-                for (int j = 0; j < outputDim; j++)
-                {
-                    flattened[(i * outputDim) + j] = j < rows[i].Length ? rows[i][j] : NumOps.Zero;
-                }
-            }
-
-            return (TOutput)(object)flattened;
+            // A flat Vector<T> has no row boundary. Although all components could be copied into it,
+            // the meta-learning consumers treat a vector as one prediction: ComputeAccuracy takes one
+            // global argmax and ComputeLossFromOutput passes it directly to the configured loss. Until
+            // those consumers become OutputDimension-aware, returning row-major data here would make a
+            // valid representation locally and give it the wrong meaning everywhere downstream.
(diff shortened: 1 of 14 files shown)


ooples:
Superseded by #2029. Every commit on `fix/n1-quantum-nan` has been merged into `fix/n1-nonfinite-training`, so the whole N1 cluster now lives on a single branch and PR:

- `fix(quantum): encode a valid normalized state instead of sqrt of raw features`
- `fix(mbpa): return the k nearest in order and stop refusing batched multi-component output`
- `fix(tests): well-pose the memorization target for born-rule heads`
- `fix(n1): address quantum and MbPA review findings`
- `fix(n1): address follow-up review edge cases`
- `fix(n1): stabilise cutie memory readout and repair non-finite comparisons`
- `fix(lstmvae): implement the elbo backprop and stop forecasting from an anomaly score`

Splitting this out was my mistake — the work was meant to continue on the existing N1 branch. Closing here; review continues on #2029.

PR #2037, opened 2026-08-22
fix(som): Predict returned its own input instead of a BMU one-hot

ooples:
## The bug

`SelfOrganizingMap.PredictCore` opened with `TryForwardGpuOptimized`, which delegates to
`ForwardGpu` and walks the **layer chain**. This model deliberately has none — `InitializeLayers`
is empty because a SOM's parameters are its codebook in `_weights`, not a stack of layers.
Walking an empty chain returns the input verbatim, so with a `DirectGpuTensorEngine` (the default
here) `Predict` returned its own 128-length input rather than a one-hot activation over the map.
**The BMU was never computed**, and every consumer of the documented one-hot contract — clustering,
quantization error — silently received a passthrough.

Measured on a default `SelfOrganizingMap<float>`:

| | before | after |
|---|---|---|
| output length | 128 (the input) | **64** (the map) |
| nonzeros | 128 | **1** — one-hot |
| BMU for x vs 10x | n/a (identity) | 34 vs 55 |

## The premise was wrong, and that is the interesting part

This started from "SOM fails `ScaledInput_ShouldChangeOutput` with all-zero output". Neither half
held.

- **All-zero output was never real.** The assertion printed only the first eight elements, so a
  one-hot whose winner sits at index 20 rendered as `[0,0,0,0,0,0,0,0]` and read as a dead forward
  pass. That single misleading string is what the diagnosis rested on. The message now reports
  length, nonzero counts and argmax, and names which defect it is — an entirely zero output and an
  unmoved one-hot are different failures wanting different fixes.
- **SOM was passing that test, not failing it.** On the broken path an identity passthrough
  trivially changes when the input is scaled, so the invariant was green *for the wrong reason* and
  caught nothing. Implementing what was described would have fixed nothing real.

## Exemptions, and why the order mattered

Once `Predict` produces a genuine one-hot, the test fails honestly: `len=64 nonzeros=1/1
argmax=20/20` — a valid one-hot that did not move. A SOM's forward is `argmin_n ||w_n - x||²` over
a codebook; scaling `x` by ten preserves **direction**, and for a point far outside the codebook
that argmin reduces to `argmax_n (x · w_n)` — frequently the same neuron. `RadialBasisFunctionNetwork`
is in the same structural position: `exp(-β‖x-c‖²)` underflows for both `x` and `10x` far from every
centre.

Both are exempted via a new `ScaledInputInvariantApplicable` hook following the house pattern
(`TrainingErrorInvariantApplicable` and siblings), each with its own measured rationale.

**Exempting SOM first would have buried the defect.** The passthrough made the invariant pass; the
failure only surfaced after the forward was fixed. The exemption is only defensible because the bug
behind it is fixed — which is the opposite of the order the original framing implied.

## Also corrected

`OutputShape` was `[60]`, dating from the grow-then-shrink grid heuristic that #1789 replaced with
an exact factor-pair choice — `outputSize` 64 now allocates exactly 64. It read 60 unnoticed
precisely because the passthrough meant the declared output shape was never exercised.

## Verification

SOM (neural + clustering), RBF and their families: **77 passed, 2 skipped, 0 failed.**

### DIFF
--- a/REFACTOR_STATE.md
+++ b/REFACTOR_STATE.md
@@ -1,0 +1,44 @@
+# Generator incremental-caching refactor — state
+
+Branch: refactor/generator-incremental-caching (worktree AiDotNet-genrefactor)
+Pushed: 688aa13ef
+
+## The defect being fixed
+Generators returned INamedTypeSymbol from CreateSyntaxProvider `transform:` and then
+Combine()d with CompilationProvider. ISymbol is not value-equatable (nothing caches)
+and each retained symbol roots the whole Compilation (cache pins compilations in RAM).
+That is the VBCSCompiler multi-GB growth.
+
+## Recipe (proven 4x)
+1. transform returns a value-equatable model, not a symbol; semantic work moves into Analyze().
+2. Compilation read transiently via ctx.SemanticModel.Compilation; never escapes pipeline.
+3. Collect().Combine(CompilationProvider) -> plain Collect(); Execute -> Emit(ImmutableArray<Model>).
+4. Model: immutable, ImmutableArray<>, explicit IEquatable. Equality MUST reach through
+   nested collections (nested types need IEquatable too).
+
+## Known traps
+- `.Count` on ImmutableArray binds to the LINQ extension METHOD GROUP -> use `.Length`.
+- Text anchors can collide with the generator's own emitted string literals (e.g. "return entry;").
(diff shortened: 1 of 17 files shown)


ooples:
Folding this into #2036. Splitting it into a second PR was my mistake: CI here takes hours, so two PRs means two full runs for one body of work. All commits from this branch are now on refactor/generator-incremental-caching.

PR #2058, opened 2026-08-30
refactor(optimizers): delete the cached typed options copy, 42 updateoptions overrides to 6

ooples:
> **Stacked on #2055 and depends on it.** Base it there, not on master. #2055 makes the base adopt restored options in `Deserialize`; without that, a computed property would read the *constructor's* options after a restore rather than the restored ones.

## The problem

`UpdateOptions` was `abstract`, so all **42** concrete optimizers implemented it — and **36 were the same eight lines**:

```csharp
protected override void UpdateOptions(OptimizationAlgorithmOptions<T, TInput, TOutput> options)
{
    if (options is AdamOptimizerOptions<T, TInput, TOutput> adamOptions)
    {
        _options = adamOptions;
    }
    else
    {
        throw new ArgumentException("Invalid options type. Expected AdamOptimizerOptions.");
    }
}
```

That private field is a **second copy** of state the base already owns, and keeping two copies in step is the only reason the method existed. Every new optimizer inherited the obligation, and any change to how options are held meant 42 edits.

## The change

The 36 read their typed options through a computed property over the base's single instance:

```csharp
private AdamOptimizerOptions<T, TInput, TOutput> _options
    => (AdamOptimizerOptions<T, TInput, TOutput>)Options;
```

The override then has nothing to do and is deleted. `UpdateOptions` becomes `virtual` with a no-op default, to be overridden only to **react** to a new option set.

**Six still do, and keep both field and override:** ADMM and ProximalGradientDescent rebuild a regularizer, Bayesian updates a kernel, GradientBasedOptimizerBase caches a scheduler mode, Lion re-seeds adaptive parameters, Rprop validates hyperparameters.

## It closes a latent bug, not just noise

The constructors read:

```csharp
: base(model, options ?? new())
{
    _options = options ?? new AdamOptimizerOptions<T, TInput, TOutput>();
```

When the caller passes `null` those are **two different default instances** — the base holding one, the derived class the other, each blind to changes in the other. There is now one instance and no way to desynchronise them.

`Adam8BitOptimizer`'s hand-rolled `Deserialize` kept its own copy of the options JSON too. The read stays — the stream position depends on it and a malformed payload must still fail there — but the value is discarded, since `base.Deserialize` has already adopted the authoritative copy.

## Verification

- **1693 of 1693** optimizer tests pass.
- `src` builds clean on **net10.0, net8.0 and net471**.
- **42 → 6** overrides; net **-780** lines.

## Note on how this was produced

The mechanical conversion was scripted, but two passes had to be redone: a first regex over-reached into the six exception files and stripped constructor assignments they still need, and `RootMeanSquarePropagationOptimizer` is a `partial class` whose override lives in another file, so a per-file pass missed it. Both were caught by the compiler, not by inspection — worth knowing if this pattern is applied elsewhere.

https://claude.ai/code/session_01U1HFtoj81avXvuLebTpWyk

### DIFF
--- a/src/AiDotNet.Generators/AnalyzerReleases.Unshipped.md
+++ b/src/AiDotNet.Generators/AnalyzerReleases.Unshipped.md
@@ -81,3 +81,4 @@ AIDN097 | AiDotNet.FacadeConfiguration | Warning | FacadeConfigurationValidation
 AIDN098 | AiDotNet.ParameterAutomation | Warning | TrainableParameterGenerator, Declared parameter axis cannot be proven resolved
 AIDN099 | AiDotNet.ParameterAutomation | Warning | TrainableParameterGenerator, [TrainableParameter] on a non-partial class does nothing
 AIDN046 | AiDotNet.TestCoverage | Warning | TestScaffoldGenerator, Layer cannot be scaffolded and produces no generated tests
+AIDN077 | AiDotNet.GoldenPattern | Warning | GoldenPatternValidationGenerator, Optimizer builds its own random generator instead of drawing from the seeded OptimizerBase.Random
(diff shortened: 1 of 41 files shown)


ooples:
Folded into #2055 rather than run a second multi-hour CI cycle on work that cannot build without it. The commit `c687c6fd65` is unchanged and now sits on top of the seeding fixes in that branch — nothing is lost by closing this.

Splitting them was my mistake: the refactor depends on #2055's `ApplyOptions` to adopt restored options, so the two were never independently mergeable and two PRs meant paying CI twice for one logical change.

PR #2075, opened 2026-09-01
fix(ci): group the vi/vi alternation before excluding video

ooples:
The `ModelFamily - Generated Layers Vi-VM` shard means to take `Vi*` and `VI*` while leaving
`Video*` to its own dedicated shard. It does not.

VSTest binds `&` tighter than `|`, so

```
(A | B & C) | D    parses as    (A | (B & C)) | D
```

which applies the `!~Generated.Video` exclusion only to the `VI` branch. Matching is also
case-insensitive, so `Generated.Video` enters through the `Generated.Vi` branch and the exclusion
never sees it.

**Every Video test runs twice** — once in `Vi-VM`, once in the `Video` shard.

## Measured, not reasoned

A three-class probe (`VideoTests`, `VisionTests`, `VMTests` under `ModelFamilyTests.Generated`),
listing tests per filter against real VSTest:

| filter | tests selected |
|---|---|
| `Vi-VM`, current | `VMTests`, **`VideoTests`**, `VisionTests` |
| `Vi-VM`, fixed | `VMTests`, `VisionTests` |
| `Video` shard | `VideoTests` |

The union across both shards is unchanged, so nothing stops running — the fix only stops `Video`
running twice.

Parenthesising the alternation explicitly also makes the expression correct whichever way the
precedence question is answered, rather than depending on it.

## Scope

Swept the other 115 shard filters for the same shape (an ungrouped alternation followed by an `&`
exclusion). This was the only one.

Found by CodeRabbit on #2072, where the shard definitions are being relocated; it is pre-existing
on master, so it is fixed here rather than there.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1234,7 +1234,7 @@ jobs:
             framework: net10.0
             filter: >-
               FullyQualifiedName~ModelFamilyTests.Generated&
-              ((FullyQualifiedName~Generated.Vi|FullyQualifiedName~Generated.VI&FullyQualifiedName!~Generated.Video)|FullyQualifiedName~Generated.VM)
+              (((FullyQualifiedName~Generated.Vi|FullyQualifiedName~Generated.VI)&FullyQualifiedName!~Generated.Video)|FullyQualifiedName~Generated.VM)
           - name: ModelFamily - Generated Layers Vo-VR
             project: tests/AiDotNet.Tests/AiDotNetTests.csproj
             framework: net10.0


ooples:
Closing: this should not have been a separate PR. The fix is already committed to #2072, where the shard definitions now live, and #2072 replaces this block in sonarcloud.yml entirely when it merges - so a standalone PR fixes nothing that #2072 will not, while costing a full ~2h 116-shard run that competes for the same 20 runner slots as the verification run currently in flight.

The fix and its VSTest measurements are in #2072. Apologies for the churn.

Issue #2093, opened 2026-09-05
FastSpeech2 training diverges (loss 9.66 -> 15.63) and is invisible to CI: HeavyTimeout-excluded from the gate, nightly lane red


ooples:
## Summary

`FastSpeech2` does not train — its loss **increases** — and no CI lane can see it, because all 34 of its tests are `HeavyTimeout`-tagged out of the PR gate and the nightly lane meant to catch them has been failing for five consecutive nights (#2087).

```
AiDotNet.Tests.ModelFamilyTests.Generated.FastSpeech2Tests.Training_ShouldReduceLoss [FAIL]
  Training did not reduce loss: initial=9.664454, final=15.626261.
  Gradient computation or parameter update may be broken.
```

Loss goes **9.66 -> 15.63**. "Worse with more training" is a distinct signature — it points at a sign error, a runaway penalty term, or an update applied in the wrong direction, not at undertraining.

## It is deterministic, not a flake

| Run | Result |
|---|---|
| In a 1399-test batch, Debug | FAIL |
| Isolation, Debug | FAIL — `initial=9.664454, final=15.626261` |
| Isolation, **Release** | FAIL — `initial=9.664454, final=15.626261` |

Bit-identical across Debug and Release and across isolation vs batch. This is a reproducible defect, not parallelism or numeric noise.

Dependency ruled out: this worktree resolves `AiDotNet.Tensors 0.130.2` from `nuget.org` (`.nupkg.metadata` confirms `"source": "https://api.nuget.org/v3/index.json"`), the same package CI restores. It is not a local-build artifact.

## Why CI is green anyway

The shard filter matches it. `.github/test-shards.yml` for `ModelFamily - Generated Layers F`:

```
FullyQualifiedName~ModelFamilyTests.Generated&FullyQualifiedName~Generated.F
```

`AiDotNet.Tests.ModelFamilyTests.Generated.FastSpeech2Tests` matches both clauses. But `sonarcloud.yml:1302` appends the category exclusions, and that is what removes it:

| Filter | FastSpeech2 tests remaining |
|---|---:|
| shard filter alone | **34** |
| `& Category!=ModelPerformanceCensus` | 33 |
| `& Category!=HeavyTimeout` | **0** |

So the entire class is `HeavyTimeout`, and the run `33938630584` ledger records `ModelFamily - Generated Layers F` as `total=1710, executed=1710, failed=0` — accurate for what it ran, and blind to FastSpeech2.

[...]

PR #2097, opened 2026-09-06
feat(optimizers): let models declare the optimizer settings their paper specifies (#1928 mechanism)

ooples:
Refs #1928 — the mechanism half. **No model is populated, so behaviour is unchanged.**

## The problem

685 optimizer constructions across 592 files are `optimizer ?? new XOptimizer<...>(this)`, supplying no options at all. Those models silently inherit the optimizer class's own defaults.

It is not only the learning rate. `AdamWOptimizerOptions` also contributes `WeightDecay = 0.01`, `Beta1 = 0.9`, `Beta2 = 0.999`, `Epsilon = 1e-8`. Commit `1972a510a` is the proof that bites — `SpanBasedNERBase`

> "silently took AdamW's own default decoupled weight decay of 0.01 and applied it to every parameter on every step. Yu et al. 2020 (arXiv:2005.07150) Table 1 specifies plain Adam at 1e-3 ... no span-NER paper asks for that decay."

An LR-only fix would have left that class of defect behind.

## Why this needs no per-model edits

Every one of the 685 sites **already passes the model**. So resolution can happen centrally: `PaperOptimizerDefaults.Resolve(...)` is called from the four optimizer constructors that account for all 685 sites (`AdamW` 432, `Adam` 247, `LBFGS` 5, `Adam8Bit` 1). **11 files changed, not 592.**

## Why an attribute rather than a central table

```csharp
[ResearchPaper("InternImage: ...", "https://arxiv.org/abs/2211.05778", Year = 2023, Authors = "Wang et al.")]
[PaperOptimizer(OptimizerKind.AdamW, LearningRate = 1e-4, WeightDecay = 0.05,
                Source = "Sec. 4.1, Table 8")]
```

The citation and the numbers it produced sit adjacent and can be checked against each other. A lookup table divorced from the model is precisely the artefact that rots — this whole audit began because a status list had drifted from the code it described. It also matches how the codebase already declares model facts (`[ModelDomain]`, `[ModelTask]`, `[ModelComplexity]`, `[ResearchPaper]`).

## Design points that carry weight

- **`Source` is the anti-fabrication guard.** `AIDN102` makes it an **error** to declare values without citing where they came from. An absent value falls back to a documented library default and is honest; an invented one looks authoritative and never gets re-checked.
- **Unset is `NaN`, not zero.** So a paper specifying plain Adam can declare `WeightDecay = 0` and have it stick. Encoding unset as zero would silently drop the single case that matters most — which is exactly the `SpanBasedNERBase` defect. There is a test for this.
- **Caller options always win.** `ConfigureOptimizer` is untouched; the paper is the default, not the law.
- **A declaration for a different optimizer is ignored, not applied.** A learning rate chosen for SGD is not a learning rate for Adam; transplanting across optimizers would be worse than the default it replaced.
- **Variants repeat the attribute**, keyed via `IPaperOptimizerVariant`, because papers give different rates per size (InternImage-T vs -H). An unkeyed entry is the fallback, so partial population is safe rather than half-broken.
- **Scalars only.** Warm-up and cosine decay are not expressible as attribute arguments and stay in code, where warm-up-and-decay is already the default (`cba6f558f`).

## AIDN101, and why Info

Reports models that cite a paper **and** construct an optimizer with no options — **559 today**.

Two calibration decisions, both measured rather than guessed:

- **Info, not Warning.** `WarningsAsErrors` is on for this project: at Warning this rule produced **`error AIDN101` x 559** and failed the build. That volume also gets suppressed wholesale rather than worked down. This follows the ladder `AIDN087` already documents — report while the backlog is large, promote as it shrinks.
- **Both conditions required.** Flagging on `[ResearchPaper]` alone reported **2098** types, including analytic models that never construct an optimizer and so have no training defaults to get wrong. Requiring an actual no-options optimizer construction narrows it to the 559 that genuinely have the defect.

## Tests

`PaperOptimizerDefaultsTests` — 11 assertions, exercising the same call the optimizer constructors make:

- a model with no declaration keeps library defaults (**control arm** — without it every other assertion could pass by the mechanism simply overwriting everything)
- declared values are applied, and are asserted to *differ* from the defaults
- caller-supplied options win, by reference
- a declaration for another optimizer is not applied
- **explicit `WeightDecay = 0` is applied rather than read as unset**
- variant rows select correctly; an unmatched variant and a null variant both fall back to the unkeyed row
- a null model is safe
- an options type lacking a declared knob skips it rather than throwing

## Verification

- Builds clean on **net10.0 and net471**; `AIDN101` at Info does not affect the build.
- New tests: **11/11 pass**.
- **Behaviour unchanged:** the 1399-test batch reports **1359 passed / 3 failed / 37 skipped**, identical to baseline. The 3 failures are the pre-existing FastSpeech2 divergence (#2093).

## What is deliberately not here

**No model is populated.** Filling in 685 models' hyperparameters is research — reading papers and citing sections — not refactoring, and it cannot be done by pattern-matching. Landing the mechanism first makes the backlog countable and lets population proceed incrementally, family by family, with every entry carrying a `Source`.

## Summary by CodeRabbit

* **New Features**
  * Added paper-based optimizer configuration for models, including optimizer types, hyperparameters, learning-rate schedules, and gradient clipping.
  * Added support for selecting optimizer recipes by model variant.
  * Added analyzer diagnostics for missing paper settings, uncited hyperparameters, duplicate recipes, and unused declarations.
  * Added support for automatically creating optimizers from declared paper recipes.

* **Tests**
  * Added coverage for recipe discovery, variant selection, defaults, schedules, and explicit parameter values.


ooples:
The remaining `Model Performance Coverage and Regressions` failure is not a PaperOptimizer regression. The exact artifact shows `EchoStateNetworkTests.trainStepMs` at 83.1867 ms, while this same AMD EPYC 7763 environment had already measured 83.0062 ms in the seven-run history. The gate treated the 0.1805 ms envelope-edge difference as a hard failure despite its 25 ms timing noise floor.

Root fix: #2104. Replaying this run's exact 862 fixture results and 12 history artifacts through that fix changes the outcome from 1 error / 22 warnings to 0 errors / 23 warnings; the ESN high mode remains visible as a warning. After #2104 merges, this branch can update from master and rerun only the failed check.

PR #2098, opened 2026-09-06
feat(models): train models the way their papers say, and report when they do not (#1928)

ooples:
Closes part of #1928. Models that are given no optimizer now train the way their paper says, and report whether they did.

## The defect

685 construction sites build an optimizer with a generic library default. A model whose paper specifies SGD at 0.1 was training as Adam at 0.001, and nothing anywhere said so. The recipe existed only in prose, if at all.

## What a declaration looks like

```csharp
[PaperOptimizer(OptimizerKind.SgdMomentum, LearningRate = 0.1, Momentum = 0.9, WeightDecay = 1e-4,
                Schedule = LearningRateSchedulerType.ReduceOnPlateau, DecayRate = 0.1,
                ReferenceBatchSize = 256,
                Source = "He et al. 2016, Sec. 3.4: SGD, lr 0.1 divided by 10 on plateau, ...")]
```

`Source` is required whenever values are declared and enforced by an analyzer. It is the anti-fabrication guard: a number that cannot be pointed at a section should not be declared, because an invented value looks exactly like a cited one and nobody re-checks it.

`NaN` marks "the paper does not say", so a deliberate `WeightDecay = 0` for plain Adam stays distinguishable from silence — the case that matters most, since AdamW otherwise adds 0.01 of decay to every parameter on every step.

## Reporting

`result.GetTrainingRecipe()` returns what the paper specifies, what was applied, and every difference, summarised as `Exact`, `Adapted`, `Deviated` or `NotDeclared`.

`Adapted` is not a lesser `Exact`. A rate stated for batch 4096 is not the same instruction at batch 32, so adjusting it is more faithful than transplanting it. But they are different claims, so `TrainsExactlyAsPublished()` counts only `Exact`. Every adjustment names the published rule that justifies it.

## A bug this found in itself

Declaring wav2vec 2.0's warmup **broke its training**, and the generated `Training_ShouldChangeParameters` caught it: "Parameters did not change after training. Gradients may be zero or learning rate is 0."

Two silent holes. A decay needs a horizon — without one `decaySteps` goes negative, every post-warmup step returns the floor, and the rate is zero for the whole run while the report still says `Exact`. And the ramp itself started at zero, because `step / warmup` is 0 on step 0; the reference implementations index from one. Both fixed in the code, not the test — the test was right, the learning rate really was zero.

Worth stressing: **every recipe unit test passed while training was completely dead.** The generated per-model training tests are what caught it, so those now run after each declaration batch.

## Three things the work turned up

**The linear scaling rule was being applied to every optimizer.** Goyal et al. 2017 establish it for SGD with momentum; Krizhevsky 2014 Sec. 5 derives sqrt(k) and reports k worked better in his experiments — also SGD. Neither covers Adam-family optimizers, whose per-parameter second-moment normalisation is what the derivation assumes away. Scaling AdamW's rate and citing Goyal would have been a fabricated rule wearing a real citation, and invisible, because the report would have named a genuine paper for a claim it does not make. Scaling is now gated to SGD; otherwise the paper's rate stands and the batch mismatch is reported as a caution.

**Schedules were mostly unreachable.** Three of thirteen schedulers were mapped, so ReduceOnPlateau — how a generation of vision papers state their schedule — reported as a deviation despite existing. Now mapped, plus two new ones: tri-stage (wav2vec 2.0 and much of the speech literature) and Noam (the transformer family). Warmup is composed in front of schedules that have no ramp of their own, which it previously dropped in silence.

**Some models were already right.** PANNs deliberately disables the clipping its options default to, because Kong et al. do not clip; Transformer derives its Noam peak from the architecture's own model dimension. Routing those through the factory would have made them *less* faithful while reporting the opposite. `VerifyHandBuilt` keeps the model's own optimizer and turns the declaration into an assertion on it — strictly stronger than the factory path, where the recipe is the configuration and cannot disagree with itself.

## What is deliberately not declared

Roughly as much work went into refusing to declare as into declaring. USM names two optimizers but no rates. Zipformer uses ScaledAdam and the Eden schedule, and declaring Adam would contradict its own argument. TabTransformer tunes its rate per dataset by search — a search is not a value. WavLM's fine-tuning table maps a rate to each of thirteen tasks, and neither pdftotext, its layout mode, nor a coordinate rebuild recovers the mapping unambiguously; assigning speaker verification's rate to emotion recognition would be indistinguishable from a fact.

`NotDeclared` is an honest report. A guess is not.

## Method

Every value was read out of the paper PDF, not recalled and not taken from a fetch summary. This matters: Whisper's Appendix F is absent from the ar5iv rendering, so summarising the page reports "not stated" for the entire hyperparameter table.

## Scope

62 model files declared here of 852 that construct an optimizer, kept under the 100-file limit. A durable queue tracks the remainder with per-paper status, so the sweep resumes rather than restarts.

Verified: 38 recipe and scheduler tests plus 44 generated per-model training tests pass; the generated training tests for every declared family pass (the one exception, FastSpeech2, fails identically with the declaration reverted, so it is pre-existing on master); net10.0, net471 and the analyzer project all build clean.


ooples:
## Triage of the three failing shards

Reproduced locally on this PR's head (`0279a3324`) in an isolated worktree. The three failures split into two different things.

### Flaky, not caused by this PR — Sonata + InstantStyle

- `SonataTests.Predict_ShouldBeDeterministic` — CI delta `1.4901161e-08`, which is exactly 2⁻²⁶, i.e. float32 epsilon
- `InstantStyleModelTests.Clone_ShouldProduceIdenticalOutput` — CI delta `2.24e-05` against a tolerance of `1.60e-05` (1.4× over)

Neither model declares a `[PaperOptimizer]` recipe, so this PR does not touch their training path. **Both pass 5/5 locally on this exact commit.** They sit right at the tolerance boundary and appear to depend on the runner (SIMD width / parallel reduction order), not on this change. Recommend not adjusting tolerances on a hypothesis — worth confirming once CI is scheduling normally again.

### Real, and caused by this PR — ECAPA-TDNN

`ECAPATDNNSpeakerTests.LossStrictlyDecreasesOnMemorizationTask`: loss moves `0.438444 → 0.438423` over 100 steps, so the model is effectively not training.

The cause is the declared recipe reaching training differently than intended:

```
master:  _optimizer = optimizer ?? new AdamWOptimizer(this)      // defaults, no schedule
here:    _optimizer = optimizer ?? PaperOptimizerFactory.CreateFor(this)
         [PaperOptimizer(Adam, LearningRate = 1e-3, MinLearningRate = 1e-8,
                         Schedule = Cyclic, StepSize = 65000, Triangular2)]
```

A cyclic schedule starts at its **base**, which here is `MinLearningRate = 1e-8`, and `GradientBasedOptimizerBase` sets the optimizer's rate from `scheduler.CurrentLearningRate` at construction. Investigation of whether the tape-training path advances the scheduler is in progress — that determines the right fix, and it would affect every model declaring a schedule, not just this one.

Still open; I am not proposing a change until the mechanism is confirmed.

PR #2119, opened 2026-09-08
feat(models): declare nine video super-resolution recipes (#1928 batch 4)

ooples:
Fourth batch of paper-faithful training recipes for #1928, stacked on #2117.

Declares nine video super-resolution models: IART, PSRT, FlashVSR, RealisVSR, DAM-VSR, RVRT,
TTVSR, DOVE and MIA-VSR.

## Reading notes

TTVSR declares no learning rate. The paper sets 1.25e-5 for the motion estimation and 2e-4 for the
other parts, and the model builds one optimizer over both, so neither value would be correct for
all of it; its optimizer, betas, batch size and schedule are declared as stated.

PSRT's `64` is the low-resolution patch size, not a rate. DAM-VSR's rate is written with the
multiplication sign lost in PDF extraction and reads as 8e-5.

DOVE and MIA-VSR each declare two phases. DOVE additionally states a beta3 of 0.98 for a
three-moment AdamW variant this library does not implement, so that value is recorded in the
source text rather than declared as a field that would not mean anything.

## Verification

net10.0 builds clean. The video family's generated tests report 305 passed, 1 failed.

The one failure, `MIAVSRTests.TrainingStep_ShouldDependOnTheTarget`, fails identically with and
without this declaration -- confirmed by reverting that file alone and re-running, then re-running
again with the declaration restored. It is pre-existing and not a regression from this branch.


ooples:
The compatibility failure is now reproduced and fixed at the shared test-base level. The original net10 check missed an xUnit TensorShape assertion overload that is unavailable on net471; a local net471 build reproduced the exact CS1503 before the correction. After using typed rank/dimension assertions: 11/11 tests on every declared framework (net10/net8/net471), then a successful build of the actual full compatibility test project with zero errors, followed by 11/11 from each actual net8/net471 AiDotNetTests.dll. No skips, no production changes, no generated test-file edits. Full commands and preserved before/after evidence are in .github/PR_AUDIO_CONSTRUCTOR_REVIEW_PROOF.md. Pending hosted checks are not being represented as green.

ooples:
Fixed Generated Layers A / AudioPaLMTests.LossStrictlyDecreasesOnMemorizationTask in 9f7dc4bec8. Scope: TestScaffoldGenerator only. The selected fixture is TextToSpeech.MultiModal.AudioPaLM (CollisionOwners chooses this namespace). The constructor now honors the declared Adafactor 1e-4 recipe rather than AdamW at 1e-3; the old two-step probe observed 2.412728 -> 2.409396, a decrease smaller than the unchanged 1% requirement. The generator now gives this TTS fixture the existing 15-step conservative-recipe budget. No production optimizer/model change, no threshold relaxation, and no hand-edited generated test. Adversarial diff review checked namespace selection, unchanged assertion/optimizer, and the existing 120-second shared training budget. Local verification: 37 passed, 0 failed, 1 skipped (the separately gated ModelPerformanceCensus), 24 seconds of test execution. Command: dotnet test tests/AiDotNet.Tests/AiDotNetTests.csproj -c Release -f net10.0 --filter "FullyQualifiedName~ModelFamilyTests.Generated.AudioPaLMTests|FullyQualifiedName~PaperOptimizerBatch4ConstructorTests" (isolated artifacts path). This covers the complete generated AudioPaLM class and batch-4 constructor regressions; it is not a claim that every hosted shard has passed.

PR #2120, opened 2026-09-08
feat(models): declare paper-faithful training recipes (#1928 batch 5)

ooples:
Fifth batch of paper-faithful training recipes for #1928, stacked on #2119.

Opens with four flow- and diffusion-based text-to-speech models: MaskGCT, CoMoSpeech, E2 TTS and
Matcha-TTS.

## Reading notes

CoMoSpeech states two settings in the same section. The AdamW at 1e-3 with exponential decay of
0.5 every 50k steps is the one whose sentence names the teacher model and CoMoSpeech together; the
Adam at 1e-4 belongs to a different component.

E2 TTS and Matcha-TTS state rates and schedules but never name an optimizer. Both leave the kind
unspecified rather than borrowing one from a sibling paper — E2 TTS shares authors with F5-TTS,
which does use AdamW, but that is not a claim this paper makes.

MaskGCT's inverse square root schedule with 32K warmup is declared as Noam, which is that schedule
under its usual name in this codebase.

## Verification

net10.0 builds clean and the family's generated tests report 101 passed, 0 failed.


ooples:
The compatibility failure is now reproduced and fixed at the shared test-base level. The original net10 check missed an xUnit TensorShape assertion overload that is unavailable on net471; a local net471 build reproduced the exact CS1503 before the correction. After using typed rank/dimension assertions: 11/11 tests on every declared framework (net10/net8/net471), then a successful build of the actual full compatibility test project with zero errors, followed by 11/11 from each actual net8/net471 AiDotNetTests.dll. No skips, no production changes, no generated test-file edits. Full commands and preserved before/after evidence are in .github/PR_AUDIO_CONSTRUCTOR_REVIEW_PROOF.md. Pending hosted checks are not being represented as green.

PR #2121, opened 2026-09-08
chore: make FinancialDQNAgent's exploration schedule and target sync real

ooples:
Two defects in the same agent, both of which make its learning curve noise.

## 1. The exploration schedule never ran

`SelectAction` compared against `TradingOptions.EpsilonStart` — the **initial** value — on every call:

```csharp
if (training && RandomHelper.CreateSecureRandom().NextDouble() < TradingOptions.EpsilonStart)
```

`EpsilonEnd` and `EpsilonDecay` existed on `TradingAgentOptions` the whole time and nothing read them. With the default `EpsilonStart = 1.0` the behaviour policy is **100% uniform random for an entire run**, so nothing the network learns ever influences what the agent does, and the recorded rewards measure the random policy rather than the learned one.

Now decayed per update toward the floor, matching `DoubleDQNAgent` exactly (`_epsilon = Math.Max(EpsilonEnd, _epsilon * EpsilonDecay)`) so the two agents anneal identically rather than each having their own idea of a schedule.

## 2. The target sync was a coin flip

```csharp
if (RandomHelper.CreateSecureRandom().Next(TradingOptions.TargetUpdateFrequency) == 0)
```

At the default frequency of 1000, over a few hundred updates per run, that fires roughly **0.6 times per run** — and at random moments, so two runs with the same seed diverge. "Every N updates" now means what it says: a counter and `_trainSteps % N == 0`.

The target *network* itself is already a proper clone; this is the schedule deciding when it is refreshed.

## 3. The rate is now reported

`GetTradingMetrics` publishes `"Epsilon"`, as `DoubleDQNAgent` already does. Without it, an agent that annealed and one that never explored report identical rewards and the difference lives only in a private field — which is much of why this went unnoticed.

## Why this surfaced now

A downstream platform audited its RL bake-off and found **0 of 992 evaluations eligible**. Per-agent round-trip counts, 235 runs each:

| Agent | Mean round-trips | **Max** |
|---|---:|---:|
| IQL | 20.9 | 151 |
| PPO | 17.3 | 313 |
| A2C | 15.5 | 314 |
| SAC | 6.6 | 314 |
| **DQN** | **0.7** | **1** |

DQN never completed more than a single round trip in 235 runs. With a behaviour policy that is uniformly random and a target network refreshed ~0.6 times per run, that is the expected outcome rather than a surprise.

## Tests

`tests/AiDotNet.Tests/IntegrationTests/Finance/FinancialDqnExplorationScheduleTests.cs` — 4 new:

- decay actually happens, and reaches the floor
- the floor is respected and never undershot
- `EpsilonDecay = 1.0` is honoured as "hold the rate", not mistaken for "unconfigured"
- the rate is published so the curve can be plotted

Built through `FinanceTestHelpers` / `FinanceModelTestFactory`'s own construction path, so they exercise the shape the existing suite already trusts rather than a second, divergent one.

**Teeth verified:** removing the decay line fails 3 of 4, with `epsilon did not decay: 1 -> 1`.

## Scope

Deliberately limited to `FinancialDQNAgent`. `DoubleDQNAgent` already implements both correctly and is the reference this now matches — the goal is to remove a divergence, not introduce a third convention.

https://claude.ai/code/session_017otuSGr3GdmvPoYLbiaWR2

## Summary by CodeRabbit

* **New Features**
  * Exploration now gradually decreases during training, down to the configured minimum.
  * Target-network updates occur at consistent training intervals.
  * Trading metrics now report the current exploration rate.

* **Tests**
  * Added coverage for exploration decay, minimum limits, no-decay settings, and metric reporting.

### DIFF
--- a/src/Finance/Trading/Agents/FinancialDQNAgent.cs
+++ b/src/Finance/Trading/Agents/FinancialDQNAgent.cs
@@ -61,6 +61,19 @@ public partial class FinancialDQNAgent<T> : TradingAgentBase<T>, IGradientComput
     private readonly ReplayBuffer<T> ReplayBuffer;
     private readonly NeuralNetworkArchitecture<T> _architecture;
 
+    /// <summary>
+    /// Current exploration rate, decayed toward <c>EpsilonEnd</c> as training proceeds.
+    /// </summary>
+    /// <remarks>
+    /// Held as state because an exploration SCHEDULE is state. Reading <c>EpsilonStart</c> directly on every
+    /// action — which is what this did — pins the behaviour policy at its initial value forever: with the
+    /// default 1.0 the agent acts uniformly at random for an entire run and its learning curve is noise.
+    /// </remarks>
+    private double _epsilon;
+
+    /// <summary>Training updates applied, so the target sync can be a schedule rather than a coin flip.</summary>
+    private int _trainSteps;
+
     /// <inheritdoc/>
     public override ModelOptions GetOptions() => _options;
 
@@ -109,6 +122,7 @@ public partial class FinancialDQNAgent<T> : TradingAgentBase<T>, IGradientComput
         _qNetwork = new NeuralNetwork<T>(architecture, lossFunction: TradingOptions.LossFunction ?? new MeanSquaredErrorLoss<T>());
         _targetNetwork = new NeuralNetwork<T>(architecture.CloneForModelConstruction(), lossFunction: TradingOptions.LossFunction ?? new MeanSquaredErrorLoss<T>());
         ReplayBuffer = new ReplayBuffer<T>(options.ReplayBufferSize, options.Seed);
+        _epsilon = options.EpsilonStart;
         UpdateTargetNetwork();
     }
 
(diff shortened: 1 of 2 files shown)


ooples:
Folded into #2100, which fixes the same two defects this PR fixes — the epsilon schedule and the deterministic target sync — and does so better: it advances the **inherited** `TrainingSteps` counter that every other agent maintains and the state generator serialises, rather than adding a private counter beside it.

I opened this without checking for an existing PR against the same file. That was avoidable duplication of my own earlier work, and combining is the right outcome.

What was genuinely additive here has been carried across to #2100 in b8f8a5a:

- `GetTradingMetrics` now publishes `"Epsilon"`. The `Epsilon` property serves anything holding the concrete agent, but a harness collecting metrics *generically* — one row per agent, which is how this defect was found downstream — sees only the dictionary, and there an agent that annealed and one that never explored are indistinguishable.
- Two test cases #2100 lacked: `EpsilonDecay = 1.0` honoured as "hold the rate", and the published value matching the property.

Closing in favour of #2100.

Issue #2090, opened 2026-09-05
Epic: model Options classes expose no hyperparameters — 285 of 1,670 are empty


ooples:
## Summary

**285 of 1,670 `*Options` classes (17%) declare zero properties.** They are empty shells inheriting a
near-empty base, so a model's own hyperparameters — its width, depth, vocabulary, latent size — are not
configurable through the object users reach for.

This was found while repairing the documentation examples: 214 example assignments across **93 Options
types** set properties that do not exist anywhere in the hierarchy. They are not typos — a fuzzy match
against each type's real properties, inherited ones included, found nothing close. The examples were
describing a configuration surface that was never implemented.

`ACGANOptions` is the shape of it:

```csharp
public class ACGANOptions : NeuralNetworkOptions
{
}
```

while its documented example configured `LatentSize` and `NumClasses`.

## What users were told they could configure

The full list is committed at `docs/MISSING_MODEL_OPTIONS.tsv` (214 rows: options type, property, source
file). By frequency:

| property | occurrences |
|---|---|
| `InputSize` | 24 |
| `NumLayers` | 19 |
| `HiddenSize` | 18 |
| `VocabSize` | 15 |
| `ModelDim` | 13 |
| `ImageSize` | 13 |
| `EmbeddingDim` | 12 |
| `LatentSize` | 10 |
| `NumHeads` | 5 |

Not exotic knobs — the core dimensional surface of a neural model.

## The lever that already exists

| mechanism | coverage |
|---|---|
| `[ResearchPaper]` attribute citing each model's source paper | **2,142 models** |
| Paper-fidelity contract test asserting defaults match the published configuration | **2 models** |

`CensusPaperFidelityContractTests` already asserts SVTR's default topology layer by layer against its
released configuration. The mechanism works and is 0.1% adopted. Every model that cites a paper has a
specification for what its options should be and what their defaults should equal.

## Agreed design

**Where options live — both, with forwarding.** `Options` mirrors the dimensional settings and writes
through to the `NeuralNetworkArchitecture` it builds, so either path works. Users configure one object;
`InputSize`/`ImageSize`/`InputChannels` stay available on the architecture for callers already using it.
Two places to keep consistent, so this needs a test pinning the forwarding.

**How the surface is decided — paper-derived, enforced by fidelity tests.** For each model, take the
hyperparameters from its `[ResearchPaper]` citation, expose them on Options with the paper's values as
defaults, and add a contract test asserting the defaults reproduce the published configuration.

[...]

PR #2133, opened 2026-09-09
feat(#2090): phase 4 — embedding, retrieval and GAN families

ooples:
Phase 4 of #2090. Follows #2130 (phase 3). Spec in #2122.

**Ratchet: 875 → 741.** Three groups land here: enum/string parameters that earlier phases skipped, the 11 embedding and retrieval models, and the 10 GANs.

| step | models | params | ratchet |
|---|---:|---:|---:|
| enum/string parameters + `VisionMambaModel` | — | 14 | 875 → 861 |
| embedding & retrieval | 11 | 77 | 861 → 782 |
| GAN family | 10 | 41 | 782 → **741** |

## `PoolingStrategy` is promoted out of a generic type

This is the collision phase 1 deliberately deferred. `PoolingStrategy` was nested inside `TransformerEmbeddingNetwork<T>`, and **a type nested in a generic class is a distinct type per type argument** — `TransformerEmbeddingNetwork<float>.PoolingStrategy` and the `<double>` one were unrelated types, and neither could be named from a non-generic options class. It was never usable as configuration, which is exactly what it describes.

It becomes `AiDotNet.Enums.EmbeddingPoolingStrategy`. Renamed rather than moved as-is, because `src/VisionLanguage/Encoders` has its own unrelated `PoolingStrategy` with different members.

## Options inheritance now mirrors model inheritance

Seven embedding models derive from `TransformerEmbeddingNetwork<T>` and forward through `: base(...)`. Their options classes had to derive from `TransformerEmbeddingOptions` rather than straight from `EmbeddingModelOptions` — otherwise a derived model cannot pass its own options to base at all.

## Three assumptions the GANs broke

1. **Not every model has an architecture parameter.** The GAN primary constructors take explicit image dimensions and build their generator and discriminator architectures from them.
2. **A constructor may delegate with `: this(...)`**, not only `: base(...)`, and such a constructor is public and carries movable parameters of its own.
3. **A moved value may be used *inside* the initializer**: `: base(CreateDCGANGeneratorArchitecture(latentSize, ..., generatorFeatureMaps), ...)`. Dropping the identifier is wrong — the expression needs the value. `options` is in scope in an initializer, so it becomes `(options?.GeneratorChannels ?? 64)`, carrying the same default. And a name defaulted in one constructor but *required* in another (DCGAN's `latentSize`) must be supplied there, not dropped.

## What the tests caught that the build did not

`GanOptions.ValidateCore` required `LatentSize` and `InitialLearningRate`. **30 GAN tests failed.** `LatentSize` stays a required *constructor* argument on the four models that build architectures from it, so it is legitimately unset on those options objects; and most GANs never had an `InitialLearningRate` parameter at all. It now carries the DCGAN paper's 0.0002 as a default, and only `ImageChannels` is required.

Worth noting against the ratchet itself: it counts a parameter as covered when the options class has a matching **name**, so a model whose constructor still takes the parameter can score as migrated. `UnifiedMultimodalNetwork` did exactly that in phase 3. The skipped behavioural assertion is what closes that hole.

## Verification

- `dotnet build src/AiDotNet.csproj -f net8.0` — 0 errors
- `dotnet build tests/AiDotNet.Tests -f net8.0` — 0 errors
- GAN suite — **57 passed, 0 failed**; embedding suite — 37 passed
- Ratchet — passing at **741**

Refs #2090

### DIFF
--- a/CI_SHARD_INVENTORY.md
+++ b/CI_SHARD_INVENTORY.md
@@ -1,0 +1,119 @@
+# AiDotNet CI Shard Failure Inventory & Work-Division Plan
+
+**Source of truth:** clean `Build & SonarCloud` run on `master` @ `01a4ddb4` (Tensors **0.102.12**), run id `28023414937`.
+**Result:** **44 pass / 19 fail / 1 cancelled** (`NeuralNetworks T-Z`, runner lost) of 64 test shards.
+**Generated:** 2026-06-23. Re-run the run and re-classify before locking assignments if master has moved.
+
+> **Already fixed (do not re-investigate):** the fp32 small-M GEMM `ArrayPool` host-crash that was aborting every CNN/diffusion shard wholesale — fixed in **Tensors #672 → 0.102.12**, on master via #1667. Verified locally: `ResNetNetworkTests` 56/56, `ConvolutionalNeuralNetworkTests` 21/21.
+
+---
+
+## TL;DR — the 19 failures are 4 distinct causes, not 19 model bugs
+
+| # | Root cause | Shards | One fix greens many? |
+|---|-----------|:------:|----------------------|
+| ① | **Stack overflow** (host crash, recursion) | 4 (+1 cancelled) | Likely yes — probably one shared recursion |
+| ② | **Inference arena corruption** (`#1661`, default-ON) | 3 (+~7 in ③) | **Yes — biggest lever** |
+| ③ | **Host killed mid-run** (arena-crash *or* OOM; log truncated) | 8 | Partly (triage → ② or ④/OOM) |
+| ④ | **Genuine per-model bugs** (training / clone / construction) | 4 | No — individual |
+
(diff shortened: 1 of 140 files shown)


ooples:
Closing as superseded, with the evidence rather than an assumption.

`git merge-base --is-ancestor feature/options-surface-phase-4-embed-gan feature/options-surface-phase-7-audio` succeeds: **every commit on this branch is already contained in the phase-7 branch**, which is open as #2158 and now ready for review. Nothing here is lost by closing.

Two things worth noting about why this looked worse than it was:

- The "27 of 32 checks failing" is misleading. The run (`34361949155`, 2026-09-09) has conclusion **`cancelled`**, not `failure` — an early job failed and the rest were cancelled, so most of those 27 never actually ran and never actually failed.
- The branch is 48 commits behind `master` and its tip is 11 days old, so even a fresh run would not have told us much about the current state of the work.

The branch is not deleted. If anything here turns out not to have made it into #2158 after all, it can be recovered from `feature/options-surface-phase-4-embed-gan` at its tip.

PR #2205, opened 2026-09-15
fix(ci): accept absent optional delta map inputs

ooples:
Follow-up to #2197, discovered in actual master push 34919349139. With no certified map, the workflow supplies empty MapFile and ShardManifestFile arguments. PowerShell rejected these before exact-tree reuse or pending deferral could execute, triggering full-validation fallback. Explicitly allow empty optional inputs in Resolve mode; retain an explicit nonempty requirement in PlanDelta mode. No certificate eligibility, delta selection, or test assertions are weakened. Before: invoking the merged resolver with empty map arguments reproduces the binding error. After: 4 real script-entry cases pass (both empty, each individually empty, both omitted); Test-ValidationReuseModes passes, including 16 reconciliation cases. Test-TestImpactEndToEnd passes, including 64 eligibility cases, 50 unsafe workflow mutations, 12 production emission cases, and real-Git exact/delta routing. Diff reviewed adversarially: missing maps still cannot authorize delta reuse. Master map restoration is separately complete in successful audit 34919418484; this correction protects future absent/revoked-map states. Hosted verification pending; kept draft for review.


ooples:
Verification update: Test-ValidationReuseModes.ps1 passed (4 actual missing-map binder cases, 3 strict offline PlanDelta rejection cases, certificate/reuse/import policies, and 16 deferred-resumer cases). Test-TestImpactEndToEnd.ps1 passed (64 eligibility cases, 50 rejected unsafe workflow mutations, 12 production emission cases, and real Git delta fixtures). Adversarial review checked that optional binding does not relax mandatory planning and trigger diagnostics use environment variables instead of interpolating untrusted values into shell code. Live run 34998119531 confirms the binding error is fixed but is NOT validation proof: its captured event was synchronize with draft=true, so downstream checks skipped. Fresh validation was triggered after ready status was established. Acceptance still requires actual build/test execution and certified merge reuse; skipped checks are not proof. Normal merge was refused by repository rules; no bypass used.

PR #2204, opened 2026-09-15
docs(activation): clarify temperature conversion

ooples:
Production CI-routing acceptance canary after #2197. One inline comment changes in a mapped source file; executable code is unchanged. Uses the ordinary master-certified map with no workflow or trust overrides. Master audit 34919418484 passed and certified the map with zero historical misses. Expected shard matrix: Integration A-B plus mandatory Integration D, 2/116. Actual selection and merge-reuse results are pending. Created ready to exercise the genuine PR event rather than draft-skipped CI. Integration D runtime diagnosis remains separate in existing PRs #2183/#2112.

## Summary by CodeRabbit

- **Documentation**
  - Clarified the constructor comment for the temperature setting.
  - No runtime behavior or public functionality changed.

PR #2217, opened 2026-09-17
fix(ci): run an unmapped ordinary shard instead of escalating every pr

ooples:
## Problem

Every pull request has run the full matrix (209 shards) since #2173 merged, with:

```
the shard map ... could not be trusted - running the full matrix
(The map is missing ordinary workload 'Sweep - ParameterCountContractTests 0/8'; auxiliary rollout cannot repair that.)
```

`Complete-CiMapWorkloads` threw whenever the manifest named an ordinary shard the certified map had never measured. Selection caught the throw and escalated the whole PR. The nightly certification replays through the same function, so the map could not catch up either. The manifest grew 116 → 161 → 164 → 209 between 09-15 and 09-17.

Measured over 71 PR runs since 09-08: 13 reduced, and 7 of those were canary/proof branches. Since 09-16, every PR that reached a decision escalated.

## Change

An unmeasured workload of any kind now joins `alwaysRun`: it runs on every selection until a map includes it, and no coverage is invented for it. Auxiliary workloads were already handled this way, and `Select-Shards.ps1` already had `Add-UnmappedShardsAsAlwaysRun` for the same purpose. The throw pre-empted it. Malformed kinds and map shards the manifest no longer has are still rejected.

## Evidence (production data, not fixtures)

The selector was replayed inside each PR's GitHub merge ref, using that tree's own `test-shards.yml` and the certified map CI uses today (run 35047041346, 116 shards). Arms were master's `tools/TestImpact` versus this branch's.

| PR | master | this branch |
|---|---|---|
| #2214 (2 files) | escalated, 209 | 99 (96 unmapped + 3) |
| #2204 (1 file) | escalated, 209 | 97 (96 + 1) |
| #2100 (5 files) | escalated, 209 | 98 (96 + 2) |
| #2174 (18 files) | escalated, 209 | 126 (96 + 30) |
| #2185 | escalated, 209 | escalated: adds a test project no map shard executes (correct) |
| #2183 | escalated, 209 | escalated: generator change + new test project (correct) |

The master arm reproduces the production message exactly. `Test-CiWorkloads.ps1` fails on master's `CiWorkloadKinds.ps1` and passes here. `Select-Shards -SelfTest`, `Test-TestImpactEndToEnd`, `Test-CiImpactWorkflow`, `Test-CiPolicyImpact`, `Test-AuxiliaryInventory`, `Test-AuxiliaryEvidence` and `Test-CiGateModes` pass locally.

## Not fixed here

The 96 always-run shards are the cost of a map that predates them. They leave only when a map is certified on the current manifest. The nightly harvest still rejects any coverage run whose shard list differs from master's (`used a different shard manifest`), so one `test-shards.yml` edit during the ~24-hour dispatch-and-harvest window delays the map another day. A follow-up PR makes the harvest accept per-shard-unchanged coverage.

This PR changes selection tooling, so its own run escalates to the full matrix by design.


ooples:
Superseded by #2220, which contains this commit (0c5b8613ae) plus the harvest fix it needs. Closing to free the shared runner queue.

PR #2225, opened 2026-09-18
feat(finance): add a weight-resolution seam to PortfolioManagerEnvironment

ooples:
## The gap

A caller cannot train a policy under the same allocation rule it will be **served** under. The environment's weight rule is baked into `ApplyAction`, and the class is `sealed`, so there is no way to substitute one.

## Why it matters

A production system frequently applies a richer allocation rule at serving time — inverse-volatility weighting, a per-asset concentration cap, net-exposure limits, portfolio-volatility targeting — that this environment does not model. When training applies one rule and serving applies another, the two disagree about **more than scale: they order the book differently.**

Measured on one such system, replaying a *single trained policy* over a held-out span:

| | |
|---|---|
| conviction order differed | **32–48 of 48 steps** |
| max per-asset weight divergence | **0.17 – 0.37** |

The name the policy ranked second was served last. That is not a rescaling a downstream risk gate can undo — it is a different portfolio, and a policy trained against a rule it is never served under has learned a different problem.

## The change

- Extract the weight decode into `protected virtual double[] ResolveTargetWeights(Vector<T> action)`. **The default body is the existing rule, moved verbatim**: non-finite becomes zero, clamp to `[-1, 1]`, scale the vector only when gross exceeds the leverage budget. No existing caller's behaviour changes.
- **Unseal** the class so the method can be overridden. Verified safe: every reference in `src/` and `tests/` is a constructor call or a doc `<see cref>`; nothing switches on the exact type.
- **Guard the override's contract** — a null or mis-sized return throws rather than silently trading a book of the wrong width, which would otherwise leave trailing assets at whatever they happened to hold.

## Tests

Seven cases in `PortfolioWeightResolutionSeamTests`:

- the default rule is unchanged — pass-through inside the budget, scaled over it, clamped beyond the per-weight range;
- a non-finite element is neutralised;
- an override genuinely **reaches position reconciliation** (asserted through `GrossExposure`, which is derived from the positions actually taken, so it cannot pass unless the override took effect);
- the same action under the **default** rule produces a materially different book — the control that stops the override test passing vacuously;
- a mis-sized override is refused.

The control asserts a **property, not a figure**: a long-only book has no short proceeds to fund it and `ExecuteTrade` caps each buy at cash on hand, so realised gross settles *below* the budget rather than on it. Pinning an exact number would assert the cash cap rather than the seam.

## Verification

`AiDotNet.csproj` and `AiDotNetTests.csproj` both build; seam tests **7/7**. Opened non-draft deliberately, so the test-CI shards actually run.

https://claude.ai/code/session_017otuSGr3GdmvPoYLbiaWR2


ooples:
## The three red gates are a runner OOM, not a test failure

Recording the diagnosis so a reviewer doesn't go hunting for a regression that isn't there.

**Causal chain, reading upward:**

- `CI Gate` → `##[error]Required job 'validation-gate' reported 'Failure' (expected Success).`
- `Validation Gate` → aggregates `test-net10-sharded`
- `test-net10-sharded` → the only failing shard is **`Tests (net10.0) - Integration D`**

So all four failures are **one** cause reported three times downstream.

**Integration D was killed by the runner, mid-run:**

```
[memdiag 21:30:42] MemAvailable: 233976 kB
  Passed …LayoutAwareDocumentTests.DocFormer_BothModalities_RouteAndStayFinite [975 ms]
[memdiag 21:30:44] MemAvailable: 290968 kB
  Passed …LayoutAwareDocumentTests.LayoutLM_Predict_ReturnsOutput [2 s]
[memdiag 21:30:46] MemAvailable: 200788 kB
  Passed …LayoutAwareDocumentTests.DocFormer_NativeConstruction_Succeeds [28 ms]
[memdiag 21:30:48] MemAvailable: 158580 kB
##[error]The runner has received a shutdown signal.
```

Available memory fell from ~996 MB to **148 MB** while running the large document/vision models (LayoutLM, DocFormer, MATCHA, Nougat, DiT). Every test that reported before the kill **passed**; the step shows `cancelled` and all downstream steps `skipped`, which is the shape of an aborted run rather than an assertion failure. The job ran under `--blame-crash` with an `Upload crash evidence` step, which exists for exactly this.

**Why this change cannot be the cause:** no test under `tests/AiDotNet.Tests/IntegrationTests/` references `PortfolioManagerEnvironment` or `ResolveTargetWeights` — the modified code is unreachable from that shard.

**Caveat on the obvious counter-check:** comparing against master is not informative here. Master carries 403 checks but **zero** Integration checks, so "master is green" says nothing about this shard.

**Action taken:** re-ran the failed jobs. If Integration D goes green on the retry, the diagnosis holds; if it dies at the same memory profile again, that is a shard-sizing issue on the runner and worth its own issue rather than a change here.

**Local verification of this branch:** full `AiDotNet.sln` builds clean across all 12 projects (0 errors), and the RL + Finance suites pass **1326, 0 failed, 2 skipped**.

https://claude.ai/code/session_017otuSGr3GdmvPoYLbiaWR2

